PR #3, opened 2015-08-17
Exposing killAllSimulators

sebv:
Not sure it is the best way, but we need it.


sebv:
ping @Jonahss @jlipps

jlipps:
seems reasonable to me!

PR #6, opened 2015-08-24
Don't use symlink for log location

imurchie:
There was a race condition between getting the sim log directory (which is a symlink) and making it. We end up not having any logs.

Use the full path to the sim log directory.


jlipps:
aha!

PR #8, opened 2015-08-31
Better sim shutdown

sebv:
Was investigating why my Simulator gets corrupted, found that there was still some service running at the time of the xcrun, with the udid in it, so added code to make sure those are stopped.

ping @imurchie @jlipps

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -8,8 +8,9 @@ import { utils as instrumentsUtil } from 'appium-instruments';
 import { killAllSimulators } from './utils.js';
 import _glob from 'glob';
 import { read } from './settings';
-import { asyncmap } from 'asyncbox';
+import { asyncmap, waitForCondition } from 'asyncbox';
 import * as settings from './settings';
+import { exec } from 'teen_process';
 
 
 const glob = B.promisify(_glob);
@@ -195,6 +196,7 @@ class SimulatorXcode6 {
 
   // TODO keep keychains
   async clean () {
+    await this.endSimulatorDaemon();
     log.info(`Cleaning simulator ${this.udid}`);
     await simctl.eraseDevice(this.udid);
   }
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-6.js
```diff
@@ -227,8 +229,38 @@ class SimulatorXcode6 {
...
+    } catch (err) {
+      log.warn('Could not stop simulator daemons, carrying on anyway!');
+    }
+    try {
+      let removeCmd = `launchctl list | grep ${this.udid} | cut -f 3 | xargs -n 1 launchctl remove`;
+      await exec('bash', ['-c', removeCmd]);
+    } catch (err) {
+      log.warn('Could not remove simulator daemons, carrying on anyway!');
+    }
+    let pid;
```
imurchie: How is this `pid` being set?I see it being used in the `catch` block but nowhere does it seem to be given a value.

sebv: Forgot to remove this, fixed.

lib/simulator-xcode-6.js
```diff
@@ -227,8 +229,37 @@ class SimulatorXcode6 {
...
+    // This prevents xcrun simctl erase hangs.
+    log.debug(`Killing any simulator daemons for ${this.udid}`);
+    try {
+      let stopCmd = `launchctl list | grep ${this.udid} | cut -f 3 | xargs -n 1 launchctl stop`;
+      await exec('bash', ['-c', stopCmd]);
+    } catch (err) {
+      log.warn('Could not stop simulator daemons, carrying on anyway!');
+    }
+    try {
+      let removeCmd = `launchctl list | grep ${this.udid} | cut -f 3 | xargs -n 1 launchctl remove`;
```
jlipps: fair amount of duplication in these magic strings; should we not just to `'${magic} stop'` and `'${magic} remove'`?

jlipps:
good observation!

PR #16, opened 2015-10-15
cleanup work for appium 1.5

Jonahss:
@imurchie

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -1,5 +1,6 @@
 import path from 'path';
 import * as simctl from 'node-simctl';
+import { getPath as getXcodePath } from 'appium-xcode';
 import log from './logger';
 import { fs, util, rimraf } from 'appium-support';
 import B from 'bluebird';
@@ -8,7 +9,7 @@ import { utils as instrumentsUtil } from 'appium-instruments';
 import { killAllSimulators, endAllSimulatorDaemons } from './utils.js';
 import _glob from 'glob';
 import { read } from './settings';
-import { asyncmap, waitForCondition } from 'asyncbox';
+import { asyncmap, waitForCondition, retryInterval } from 'asyncbox';
 import * as settings from './settings';
 import { exec } from 'teen_process';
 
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-6.js
```diff
@@ -184,18 +187,33 @@ class SimulatorXcode6 {
...
+    let checkBooted = async function () {
+      let stat = await this.stat();
+      return new B((resolve, reject) => {
+        if (stat.state === 'Booted') {
+          resolve();
+        } else {
+          reject();
+        }
+      });
+    }.bind(this);
```
jlipps: why not just `let checkBooted = async () => {` so you don't have to bind?

lib/simulator-xcode-6.js
```diff
@@ -184,18 +187,33 @@ class SimulatorXcode6 {
...
+
+    await retryInterval(parseInt(BOOTING_TIMOUT/BOOTED_POLLING_INTERVAL), BOOTED_POLLING_INTERVAL, () => {
```
jlipps: i think by our conventions this should be an async function and `return await checkBooted();`

lib/simulator-xcode-6.js
```diff
@@ -184,18 +187,33 @@ class SimulatorXcode6 {
...
+    // start simulator
+    let simulatorApp = path.resolve(await getXcodePath(), 'Applications', 'Simulator.app');
+    let args = [simulatorApp, '--args', '-CurrentDeviceUDID', this.udid];
+    log.info(`starting simulator with command: open ${args.join(' ')}`);
+    await exec('open', args, {timeout: STARTUP_TIMEOUT});
+
+    //wait for it to be 'booted'
+    let checkBooted = async function () {
+      let stat = await this.stat();
+      return new B((resolve, reject) => {
```
jlipps: what is the point of creating a promise here? this whole function can go away and below you can just have as the anonymous async function:

```js
async () => {
  let stat = await this.stat();
  if (stat.state !== 'Booted') {
    throw new Error("State wasn't booted");
  }
}
```

lib/simulator-xcode-6.js
```diff
@@ -208,13 +226,18 @@ class SimulatorXcode6 {
     log.debug('Attempting to launch and quit the simulator, to create directory structure');
     log.debug(`Will launch with Safari? ${safari}`);
 
-    await instrumentsUtil.quickLaunch(this.udid, safari ? 'com.apple.mobilesafari' : undefined);
+    if (safari) {
+      await instrumentsUtil.quickLaunch(this.udid, safari ? 'com.apple.mobilesafari' : undefined);
+    } else {
+      this.run();
```
jlipps: isn't this an async function? shouldn't it be awaited?

imurchie:
I've addressed these issues in an upcoming PR. :)

PR #20, opened 2015-10-19
Isaac cleanup

imurchie:


lib/simulator-xcode-6.js
```diff
@@ -192,20 +189,12 @@ class SimulatorXcode6 {
     log.info(`starting simulator with command: open ${args.join(' ')}`);
     await exec('open', args, {timeout: STARTUP_TIMEOUT});
 
-    //wait for it to be 'booted'
-    let checkBooted = async function () {
+    // wait for the simulator to be 'booted'
+    await retryInterval(parseInt(BOOTING_TIMOUT/BOOTED_POLLING_INTERVAL), BOOTED_POLLING_INTERVAL, async () => {
```
jlipps: don't we have a styleguide thing of parseInt should always explicitly be base 10?

jlipps: it's in my linter anyway

imurchie: Interesting. Didn't know that. https://github.com/appium/appium-ios-simulator/pull/24

jlipps: yeah it's one of those random js gotchas that we try to be on the safe side of.

PR #23, opened 2015-10-20
killAllSimulators when calling sim.shutdown()

Jonahss:
@imurchie would this break anything you know of?

Prior to this change, the sim process would get "shutdown" but the Ui would remain up and be frozen. When in this state, the simulator can't be opened anymore.


imurchie:
I don't know of anything that this would break.

PR #25, opened 2015-10-22
detect simulator start time using system log

Jonahss:
@imurchie for review.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -11,7 +11,7 @@ import { asyncmap, waitForCondition, retryInterval } from 'asyncbox';
 import * as settings from './settings';
 import { exec } from 'teen_process';
 import  xcode from 'appium-xcode';
-import tail from 'tailing-stream';
+import { tailUntil } from './tail-until.js';
 
 
 class SimulatorXcode6 {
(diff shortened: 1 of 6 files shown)


lib/simulator-xcode-6.js
```diff
@@ -178,25 +180,72 @@ class SimulatorXcode6 {
     return _.compact(existence).length !== files.length;
   }
 
+  async getBootedIndicatorString () {
+    let indicator = '';
```
imurchie: This can just be a declaration, since there is no case that will not be assigned in the switch statement.

imurchie:
One comment, which doesn't need to be acted upon. Otherwise, looks good!

lib/simulator-xcode-6.js
Jonahss: :+1:

Jonahss:
@imurchie fixed this PR. using a better method to tail things instead of tailing-stream

imurchie:
Interesting. Pretty cool.

Jonahss:
@imurchie hopefully thats the last change.

lib/tail-until.js
```diff
@@ -0,0 +1,25 @@
+import { SubProcess } from 'teen_process';
+import B from 'bluebird';
+
+// tails a file, promise resolves when input string is written to file
+async function tailUntil(filePath, until, timeout = 5000) {
```
imurchie: Space before arguments.

Jonahss:
@imurchie how's it look now?

imurchie:
Looks good!

PR #28, opened 2015-11-10
BREAKING CHANGE: add simExists() function; getSimulator(udid) throws an error if udid does not exist

Jonahss:
breaking change. Before, we would just create a Simulator object with a completely made up udid, if thats what you asked for.

I don't think that's useful though. When will we be creating simulator objects that aren't tied to actual simulators.

You can still create a fake one if calling the constructor directly. But the `getSimulator` method now checks that the sim exists. Makes sense to me. @imurchie ?

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -55,6 +55,7 @@ async function endAllSimulatorDaemons () {
 }
 
 async function simExists (udid) {
+  // see the README for github.com/appium/node-simctl for example output of getDevices()
   let devices = await getDevices();
 
   devices = _.pairs(devices).map((pair) => {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,7 +1,9 @@
 import log from './logger';
+import _ from 'lodash';
 import { exec } from 'teen_process';
 import { waitForCondition } from 'asyncbox';
 import { getVersion } from 'appium-xcode';
+import { getDevices } from 'node-simctl';
 
 
 async function killAllSimulators () {
@@ -52,4 +54,17 @@ async function endAllSimulatorDaemons () {
   log.debug('Finishing ending all simulator daemons');
 }
 
-export { killAllSimulators, endAllSimulatorDaemons };
+async function simExists (udid) {
+  let devices = await getDevices();
+
+  devices = _.pairs(devices).map((pair) => {
+    return pair[1];
+  }).reduce((a, b) => {
+    return a.concat(b);
+  }, []);
+  return !!_.find(devices, (sim) => {
+    return sim.udid === udid;
+  });
+}
+
+export { killAllSimulators, endAllSimulatorDaemons, simExists };
--- a/test/unit/simulator-specs.js
+++ b/test/unit/simulator-specs.js
@@ -12,17 +12,23 @@ import B from 'bluebird';
 import xcode from 'appium-xcode';
 
 
-let should = chai.should();
+chai.should();
 chai.use(chaiAsPromised);
+
+const UDID = devices['7.1'][0].udid;
 
 describe('simulator', () => {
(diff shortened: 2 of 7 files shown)


test/unit/utils-specs.js
```diff
@@ -72,4 +77,76 @@ describe('util', () => {
       execStub.threw().should.be.true;
     });
   });
+
+  describe('simExists', () => {
+    const devicesStub = {
+    '7.1': [
+     { name: 'iPhone 5',
```
imurchie: Indentation. Throughout.

test/unit/utils-specs.js
```diff
@@ -72,4 +77,76 @@ describe('util', () => {
...
+     let results = await B.all(existence);
+
+     for (let result of results) {
+       result.should.be.true;
+     }
+  });
+
+  it('returns false if device is not found', async () => {
+    getDevicesStub.returns(Promise.resolve(devicesStub));
+    let existence = [];
```
imurchie: why not just initialize all these when the array is initialized? It's not dynamic at all.

test/unit/simulator-specs.js
```diff
@@ -12,17 +12,23 @@ import B from 'bluebird';
 import xcode from 'appium-xcode';
 
 
-let should = chai.should();
+chai.should();
 chai.use(chaiAsPromised);
 
+const UDID = 'C09B34E5-7DCB-442E-B79C-AB6BC0357417';
```
imurchie: Where does this udid come from? Will it always be there?

lib/utils.js
```diff
@@ -52,4 +54,17 @@ async function endAllSimulatorDaemons () {
   log.debug('Finishing ending all simulator daemons');
 }
 
-export { killAllSimulators, endAllSimulatorDaemons };
+async function simExists (udid) {
+  let devices = await getDevices();
+
+  devices = _.pairs(devices).map((pair) => {
```
imurchie: There needs to be some commentary in here about what the data is we are mapping, etc, so that this chain makes sense.

test/unit/utils-specs.js
Jonahss: You mean rather than pushing them onto the array?
cool idea.

test/unit/simulator-specs.js
Jonahss: it's from the 'devices.json' file that we import. so yes will always be there. I can add a comment explaining where it came from.
Hm, actually, ill just get it from the `devices` object. nicer that way anyways.

lib/utils.js
Jonahss: the exact output of `getDevices()` along with documentation is on the readme for node-simctl. Is it ok to link to that?

Jonahss:
@imurchie updated

test/unit/simulator-specs.js
```diff
@@ -12,17 +12,23 @@ import B from 'bluebird';
 import xcode from 'appium-xcode';
 
 
-let should = chai.should();
+chai.should();
 chai.use(chaiAsPromised);
 
+const UDID = devices['7.1'][0].udid;
```
imurchie: Cool. I like that better.

imurchie:
One final little thing: Can you add a little something to the README to effect that `getSimulator()` will check the udid and throw an error if it is not available.

Jonahss:
definitely. good call.

Jonahss:
squashed into two commits

Jonahss:
ooh yeh ooh yeh. checkout dat coverage increase.

Jonahss:
@imurchie ok to merge?

imurchie:
LGTM.

PR #31, opened 2015-11-20
Add method to isolate sim

imurchie:


lib/extensions/isolate-sim.js
```diff
@@ -0,0 +1,23 @@
+import { getDevices, deleteDevice } from 'node-simctl';
+import log from '../logger';
+import _ from 'lodash';
+
+
+let extensions = {};
+
+extensions.isolateSim = async function () {
```
Jonahss: Why do we have this function anyways? can we document why it exists?
Something tells me it's not actually very effective, or is not quite them best way to solve a problem.

imurchie: https://github.com/appium/appium/commit/33e1a24a459b7131c2951c9c377aae4aa91eb8d1

PR #34, opened 2015-12-01
add isRunning() and openUrl(url)

Jonahss:
Depends on https://github.com/appium/node-simctl/pull/15

@imurchie for review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,9 +132,19 @@ Retrieve state information about the simulator. Returns an object:
 
 Query whether the simulator has been run before.
 
+`async isRunning ()`
+
+Returns `true` if the simulator is running, `false` otherwise.
+
 `async run ()`
 
 Starts the simulator without any Instruments involvement, or application running.
+
+`async openUrl (url)`
+
+*Xcode version 7 and up only*
+
+Opens the input url with safari.
 
 `async clean ()`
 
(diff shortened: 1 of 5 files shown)


test/simulator-e2e-specs.js
```diff
@@ -145,6 +145,21 @@ function runTests (deviceType) {
       stat = await sim.stat();
       stat.state.should.equal('Shutdown');
     });
+
+    it.only('should detect if a sim is running', async function () {
```
imurchie: Probably don't want this `only`

README.md
```diff
@@ -132,10 +132,18 @@ Retrieve state information about the simulator. Returns an object:
...
 
+`async isRunning ()`
+
+Returns `true` if the simulator is running, `false` otherwise.
+
 `async run ()`
 
 Starts the simulator without any Instruments involvement, or application running.
 
+`async getUrl (url)`
```
imurchie: Should this be `openUrl`? And it needs a comment about only being in Xcode 7+.

lib/simulator-xcode-7.js
```diff
@@ -47,6 +47,14 @@ class SimulatorXcode7 extends SimulatorXcode6 {
       'iPhone 6s Plus (9.2)': 'iPhone 6s Plus (9.2) [',
     };
   }
+
+  async openUrl (url) {
```
imurchie: If this is only in Xcode 7, can we have some sort of thing in the 6 class so we won't get `undefined is not a function` errors? Just throw an error that makes sense?

jlipps:
:+1: to isaac's comments

Jonahss:
updated with @imurchie's comments :)

imurchie:
:+1:

PR #38, opened 2015-12-15
Use "openUrl" to start safari in launchAndQuit() instead of using Instruments

Jonahss:
This way is less hacky, and removes a dependency from the project.

I also improved the openUrl() function so that it respects async better. It used to return as soon as it was called, now it waits a little bit. (I try to tail the system logs, but unfortunately still have a hardcoded wait :( on top of that)

@imurchie for review

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -460,8 +460,12 @@ class SimulatorXcode6 {
     return [newAppPath, appPath];
   }
 
-  async openUrl () {
-    throw new Error('Command "openUrl" not supported on Xcode version 6');
+  async openUrl (url) {
+    if (await this.isRunning()) {
+      return await simctl.openUrl(this.udid, url);
+    } else {
+      throw new Error('Tried to open a url, but the Simulator is not Booted');
+    }
   }
 
   static async _getDeviceStringPlatformVersion (platformVersion) {
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -193,10 +193,8 @@ class SimulatorXcode6 {
       case '8.1':
       case '8.2':
       case '8.3':
-        indicator = 'profiled: Service starting...';
-        break;
       case '8.4':
-        indicator = 'MC: Finished cleaning up app configuration';
+        indicator = 'profiled: Service starting...';
         break;
       case '9.0':
       case '9.1':
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -5,7 +5,6 @@ import log from './logger';
 import { fs } from 'appium-support';
 import B from 'bluebird';
 import _ from 'lodash';
-import { utils as instrumentsUtil } from 'appium-instruments';
 import { killAllSimulators, endAllSimulatorDaemons, safeRimRaf } from './utils.js';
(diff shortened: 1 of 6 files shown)


lib/simulator-xcode-6.js
```diff
@@ -306,10 +293,10 @@ class SimulatorXcode6 {
     log.debug('Attempting to launch and quit the simulator, to create directory structure');
     log.debug(`Will launch with Safari? ${safari}`);
 
+    await this.run();
+
     if (safari) {
-      await instrumentsUtil.quickLaunch(this.udid, 'com.apple.mobilesafari');
-    } else {
-      await this.run();
+      await this.openUrl('http://www.appium.io');
```
imurchie: Can we open something served by appium, so we don't depend on a working internet connection?

imurchie:
Nice! Can we get a test or two?

Jonahss:
Ok I'll look at tests.

lib/simulator-xcode-6.js
Jonahss: Hmm. Except this repo is standalone, and doesn't depend on appium.
It's ok if safari can't load the url, safari will still open and will just show "safari was not able to load the page"

imurchie: Good point. Ok. Is everything ok in async-land when the url can't load?

Jonahss:
The test I added runs launchAndQuit() while passing a `true` argument to trigger safari.

I'm not so sure any other tests I could add would be useful :(

lib/simulator-xcode-6.js
Jonahss: that's worth testing. ill check

Jonahss: Yup. Safari just says it can't load the page, and we keep going.

Jonahss:
@imurchie ok to merge?

imurchie:
Ok to merge, @Jonahss

lib/simulator-xcode-6.js
```diff
@@ -460,8 +447,40 @@ class SimulatorXcode6 {
     return [newAppPath, appPath];
   }
 
-  async openUrl () {
-    throw new Error('Command "openUrl" not supported on Xcode version 6');
+  async openUrl (url) {
+    const SAFARI_BOOTED_INDICATOR = 'MobileSafari[';
```
jlipps: i tend to use UPPER_SNAKE_CASE for file-level constants, not function-level constants. not sure if we care about that as a style thing.

lib/simulator-xcode-6.js
```diff
@@ -460,8 +447,40 @@ class SimulatorXcode6 {
...
+    const EXTRA_STARTUP_TIME = 3 * 1000;
+
+    if (await this.isRunning()) {
+      await simctl.openUrl(this.udid, url);
+      await this.tailLogsUntil(SAFARI_BOOTED_INDICATOR, SAFARI_STARTUP_TIMEOUT);
+      // So sorry, but the logs have nothing else for Safari starting.. just delay a little bit
+      log.debug(`Safari started, waiting ${EXTRA_STARTUP_TIME}ms for it to fully start`);
+      await B.delay(EXTRA_STARTUP_TIME);
+      log.debug('Done waiting for Safari');
+      return;
```
jlipps: what does this return doe?

PR #39, opened 2016-01-06
in isFresh, log which files we can't find and remove no-op existence map

jlipps:


lib/simulator-xcode-6.js
```diff
@@ -174,11 +174,14 @@ class SimulatorXcode6 {
       return path.resolve(this.getDir(), s);
     });
 
-    let existence = await asyncmap(files, async (f) => { return fs.hasAccess(f); });
-
-    existence = await B.all(existence); // will throw an error if an fs.stat call fails
+    let existences = await asyncmap(files, async (f) => { return fs.hasAccess(f); });
+    for (let i = 0; i < files.length; i++) {
```
Jonahss: rather than using an iterative forloop, this could be done using Array.forEach, which gives u value and array index

Jonahss:
uh oh. only unit tests were run in travis

jlipps:
@Jonahss how do you think we should proceed here? run e2e locally?

lib/simulator-xcode-6.js
jlipps: any advantages to using Array.forEach? I could also use `for...of` which is more our style I suppose.

Jonahss: Yeah just style, no real advantages.

Jonahss: Run locally yes. I ran the tests locally myself and they failed, so there's
work to do. Could be related to my updating to El Capitan

jlipps: ok downloading the sdks so i can run it locally without my changes to be sure the tests pass then will run with my changes.

jlipps:
published as 1.1.1

PR #40, opened 2016-01-06
small fixes and enable travis e2e tests

Jonahss:


jlipps:
:+1:

Jonahss:
these are still in progress. Not going to merge until I figure out how to get travis to pass.

SrinivasanTarget:
(y)

.travis.yml
```diff
@@ -1,8 +1,15 @@
+env:
+  - _FORCE_LOGS=1
 language:
   - objective-c
-
+osx_image: xcode7.1
 before_install:
-
+  # TavisCI's OSX environment gives us node 0.10.32
```
imurchie: Why do all this twice?

lib/simulator-xcode-6.js
```diff
@@ -447,12 +440,30 @@ class SimulatorXcode6 {
...
 
+    // extra time added before opening a url by default.
+    // override by setting to zero
+    if (bufferTime === null) {
+      let version = await this.getPlatformVersion();
+      version = parseInt(version[0]);
+      if (version === 8) {
+        bufferTime = 4 * 1000;
+      }
+      else if (version === 9) {
```
imurchie: Move up a line.

lib/utils.js
```diff
@@ -10,18 +10,33 @@ import { fs } from 'appium-support';
...
+    appName = 'Simulator';
   } else {
-    args.push('iOS Simulator');
+    appName = 'iOS Simulator';
   }
 
   try {
-    await exec('pkill', args);
-  } catch (ign) {}
+    await exec('osascript', ['-e', `quit app "${appName}"`]);
```
imurchie: To run AppleScript, do we need to ensure that there is sufficient privileges?

.travis.yml
Jonahss: oops, good catch.

lib/utils.js
Jonahss: Hm. It works on travis right now. And I didn't do anything special related to privileges.

Do you know of it not working?

imurchie: No. Just wondering. Generally you need to allow it to happen. But the Travis VMs might already have it.

You can try locally by going into "System Preferences", then "Security and Privacy", "Privacy", "Accessibility", and unchecking your terminal, or whatever is running the Appium server.

Jonahss: hm. so appium users will have an issue with this?
Should we add it to Doctor or something?

imurchie: I think we should test it and make sure that it is needed. In my experience a popup will come up and things will fail. But it might work?

Jonahss:
okay @imurchie this PR is finished. Does everything look all right?

lib/utils.js
Jonahss: I'll create a ticket to try this with a fresh install of appium on a fresh mac.

imurchie:
Looks good to me! I like the cleaned up killing logic.

test/simulator-e2e-specs.js
```diff
@@ -189,8 +202,7 @@ function runTests (deviceType) {
       sim = await getSimulator(udid);
       await sim.run();
       await sim.shutdown();
-      await B.delay(60000);
-      // if (true) throw 'heh';
+      await B.delay(4000);
```
jlipps: would like a comment for this mysterious static wait

jlipps:
:+1:  on the changes!

PR #41, opened 2016-01-07
freshSim fix + plist tweak

scottdixon:
- Resolves issue where iOS 8 and iOS 8.1 get stuck in a loop waiting for a fresh simulator.
- Takes advantage of a laxer parsePlistFile method, returning a blank plist file instead of throwing.

@jlipps

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -162,6 +162,7 @@ class SimulatorXcode6 {
     ];
     let pv = await this.getPlatformVersion();
     if (pv !== '7.1') {
+      // iOS 7.1, 8.0, 8.1 and 9+ don't have a DeviceRegistry.state file
       if (pv !== '8.0' && pv !== '8.1' && pv[0] !== '9') {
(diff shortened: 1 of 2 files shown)


imurchie:
:+1:

lib/simulator-xcode-6.js
```diff
@@ -162,7 +162,7 @@ class SimulatorXcode6 {
     ];
     let pv = await this.getPlatformVersion();
     if (pv !== '7.1') {
-      if (pv[0] !== '9') {
+      if (pv !== '8.0' && pv !== '8.1' && pv[0] !== '9') {
```
jlipps: with this kind of thing where we have some magic values encoded in a conditional, i'd love to see a comment explaining why we are doing something with just these platforms.

i realize the comment should already have existed so it's not your fault, but could you add one?

scottdixon: Good call. Updated!

jlipps:
:+1:

PR #43, opened 2016-01-12
fixed bundle path map async issue

scottdixon:
Please review @imurchie 

Async issue resolved:
image

We could eventually collapse these logs into one or two lines. To do this we could use a global logging variable or pass through an additional parameter - but this would get messy with 3 libs involved.

Maybe something like:
```
log.debug('Updated Safari plist');
env.logging = 'none'
// process our plists
env.logging = 'info'
log.debug('Updated Safari plist');
```

Is something like this already in place?

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -111,14 +111,16 @@ class SimulatorXcode6 {
       };
     }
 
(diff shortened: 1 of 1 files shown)


imurchie:
I don't think anything like that exists.

But this is great! Good catch.

imurchie:
Looks good to me!

PR #45, opened 2016-01-13
update for ios 9.3 on xcode 7.3

Jonahss:


lib/simulator-xcode-7.3.js
```diff
@@ -0,0 +1,32 @@
...
+      'Library/Logs',
+      'Library/Preferences/.GlobalPreferences.plist',
+      'Library/Preferences/com.apple.springboard.plist',
+      'var/run/syslog.pid'
+    ];
+  }
+
+  async buildBundlePathMap (subDir = 'Containers' + path.sep + 'Data') {
+    if (subDir === 'Data') {
+      subDir = 'Containers' + path.sep + 'Data';
```
imurchie: String interpolation?

lib/settings.js
```diff
@@ -35,9 +36,15 @@ async function plistPaths (sim, identifier) {
       paths.push(path.resolve(sim.getDir(), 'Library', 'Preferences', 'com.apple.locationd.plist'));
       break;
     case 'userSettings':
-      paths.push(path.resolve(sim.getDir(), 'Library', 'ConfigurationProfiles', 'UserSettings.plist'));
-      paths.push(path.resolve(sim.getDir(), 'Library', 'ConfigurationProfiles', 'EffectiveUserSettings.plist'));
-      paths.push(path.resolve(sim.getDir(), 'Library', 'ConfigurationProfiles', 'PublicInfo', 'PublicEffectiveUserSettings.plist'));
+      if (compare(sim.xcodeVersion.versionString, '7.3') < 0) {
```
imurchie: It might worthwhile to get the data directory once and then use a local variable. Seems like a lot of `sim.getDir()` (though I know that has always been there).

scottdixon:
Couple of complaints from JShint otherwise LGTM!

Jonahss:
Yeah saw those :)
Fixing per isaac's recommendations.

lib/settings.js
Jonahss: I think this sort of optimization is performed by the compiler. I'm not so sure making the change improves readability either.

Made the change anyways since it's minor :)

PR #46, opened 2016-01-21
Remove logs from isFresh check

imurchie:
We don't need to have the logs folder present to consider the file not fresh.


Jonahss:
simple enough :)
:+1:

PR #47, opened 2016-01-22
added create device method

rafael-chavez:
Related to https://github.com/appium/appium-ios-driver/pull/79

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -368,6 +368,10 @@ class SimulatorXcode6 {
     await simctl.deleteDevice(this.udid);
   }
 
+  async create (deviceName, deviceType, sdkVersion) {
+    await simctl.createDevice(deviceName, deviceType, sdkVersion);
+  }
+
   async updateSettings (plist, updates) {
     return await settings.updateSettings(this, plist, updates);
   }


imurchie:
To me this doesn't make much sense to be on an already created sim object. It would be more logical to be something on https://github.com/appium/appium-ios-simulator/blob/master/lib/simulator.js

imurchie:
@Jonahss might have more thoughts on the matter.

Jonahss:
So appium so far has always reused existing sims on the host machine. After a test, we want to leave that sim back the way we found it, meaning we run `simctl erase`. Or, a user can optionally specify leaving the device with whatever state was imparted on it during the test session.

When working on appium-xcuitest-driver I reimagined the sim startup process. Taking advantage of the `simctl` utility, it creates a new simulator for every test session, deleting it afterwards. A user can specify different ways of leaving the sim intact or can specify running on an already-created simulator. I like the new default of creating a fresh sim for each test. It keeps things cleaner and doesn't require that the host machine have the default Xcode sims installed and named a certain way. 

As regards #79, appium 1.5 shouldn't be deleting any simulators, and shouldn't be creating any either.

I hope to bring the new creation/deletion reset logic to the appium-ios-driver but this would be a breaking change, hence a new feature which we can add after 1.5. For version 1.5 we want there to be as little breaking changes as possible in order to get everyone onto the newest version.

What does everyone else think?

Jonahss:
P.S. I also look forward to automatically creating/launching android emulators too.

rafael-chavez:
@Jonahss So in the beginning of the test run, Appium 1.5 is deleting the simulator that I am specifying for the run, thats the reason I added this. You can see in https://github.com/appium/appium-ios-driver/pull/79/files#diff-f64f06d264dffd32eb8d87f1b3095989R481 that on a full reset, the deleteSim() method is called, but we the flow never triggers the simulator to get created (this is when the fullReset flag is set). On full reset we are either not supposed to be deleting the simulator (what you mentioned in your comment), or if we are, we should be re-creating the simulator we just deleted. Do you agree?

rafael-chavez:
@Jonahss @sebv ^^^^^^^^

imurchie:
Closing, in lieu of cleaning, and a future new strategy for creating sims.

PR #48, opened 2016-01-22
Make sure locale is not updated if it is not necessary

imurchie:
Updating the locale is a slow process (requiring restart of the device, for instance). So only do it if there are actually changes.

Also, update `lodash`.


Jonahss:
looks good. 
Yay to updating lodash. It's been relatively painless updating it gradually this way. Plus our tests are really catching things!
:+1:

PR #57, opened 2016-05-16
Make sim check more explicit

imurchie:
Attempting to provide more information for cases like https://github.com/appium/appium/issues/6452


jlipps:
👍

PR #58, opened 2016-06-24
Adds preliminary support for Xcode 8

imurchie:
Just allow it through, and return the same driver as for 7.3. If anything comes up as different, we can fix it.


jlipps:
👍

PR #64, opened 2016-08-02
Handle boot waiting for iOS 10

imurchie:
There is no reliable boot message in the logs, so rely on the simctl status, which is not as reliable but better than waiting 60s to timeout.


jlipps:
👍

jlipps:
this sucks

PR #65, opened 2016-08-13
Add iOS 10 boot indicator

imurchie:
The latest Beta has a marker that seems to always come up.

Without this we get errors, especially in Travis, when the sim is "Booted" but not ready to be used.


jlipps:
that's exciting!

imurchie:
Argh. Unfortunately it doesn't seem to be the case in Travis.

PR #66, opened 2016-08-22
Remove file that is often not there

imurchie:
In iOS 10 the files have changed a little bit, for testing whether the sim is fully prepared.

### DIFF
--- a/test/simulator-e2e-specs.js
+++ b/test/simulator-e2e-specs.js
@@ -10,7 +10,7 @@ import getAppPath from 'sample-apps';
 import { retryInterval } from 'asyncbox';
 
 
-const LONG_TIMEOUT = 120*1000;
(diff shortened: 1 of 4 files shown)


test/simulator-e2e-specs.js
```diff
@@ -294,6 +294,13 @@ if (!process.env.TRAVIS) {
         device: 'iPhone 5s'
       }
     ];
+  } else if (process.env.DEVICE === '10' || process.env.DEVICE === '10.0') {
+    deviceTypes = [
+      {
+        version: '10.0',
+        device: 'iPhone 5s'
```
jlipps: very minor, but do you think the default device for ios 10 should be something more recent like a 6s?

imurchie: That seems reasonable, yes.

PR #69, opened 2016-10-02
fixing iFresh call in clean safari

yaphatak:


imurchie:
Published in `appium-ios-simulator@1.7.10`

PR #76, opened 2016-10-28
map bare 'iPad Simulator' to 'iPad Retina' in ios 10

jlipps:
part of appium/appium#7122

cc @imurchie


jlipps:
published as 1.7.13

PR #78, opened 2016-11-03
Update boot indicator for iOS 10

imurchie:


jlipps:
no more boot indicator for 10.1?

LGTM

PR #79, opened 2016-11-11
Use -Fn to open sim

imurchie:


jlipps:
LGTM. what does this do though?

imurchie:
It makes sure that `open` does not recycle caches and whatnot. Apparently it should help with flakiness of sim runs.

PR #80, opened 2016-11-20
Certificate installation

dpgraham:
-Can install an SSL certificate to an IOS Simulator's keychain
-Created a file called 'certificate.js' that has methods for accessing the keychains SQLITE database and that parses .pem files to create entries in said keychain
-Wrote unit tests that cover both of those classes
-Created a screen capture video that demonstrates it working: https://www.dropbox.com/s/g0dyeg0i22fleon/certificate-installation-recording.mov?dl=0
-Next step is to expose it as an API method and then to update 'appium' core project to add it to the desiredCapabilities object


lib/certificate.js
```diff
@@ -0,0 +1,162 @@
+import sqlite3 from 'sqlite3';
+import Promise from 'bluebird';
+import openssl from 'openssl-wrapper';
+import crypto from 'crypto';
+
```
jlipps: maybe these should be `const`?

lib/certificate.js
```diff
@@ -0,0 +1,162 @@
...
+let tset = `<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n
+    <!DOCTYPE plist PUBLIC \"-//Apple//DTD PLIST 1.0//EN\" \"http://www.apple.com/DTDs/PropertyList-1.0.dtd\">
+    <plist version=\"1.0\">
+    <array/>
+</plist>`;
+
+let _promiseCb = (resolve, reject) => {
```
jlipps: we use `Bluebird.promisify` instead of rolling this behavior by hand

lib/certificate.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+  };
+};
+
+/**
+ * Library for programatically adding certificates 
+ */
+class Certificate {
+
+  constructor (pemFilename){
```
jlipps: is this repo updated with new lint settings, @imurchie? should be `) {`

lib/certificate.js
```diff
@@ -0,0 +1,162 @@
...
+    let subject = await this.getSubject(this.pemFilename);
+    let trustStore = new TrustStore(dir);
+    return trustStore.removeRecord(subject);
+  }
+
+  /**
+   * Translate PEM file to DER buffer
+   */
+  async getDerData (){
+    if (this.data)
```
jlipps: we always use braces, even for one-line `if` statements. the only exception was in ES5 code (`if (err) return cb(err);`)

lib/certificate.js
```diff
@@ -0,0 +1,162 @@
...
+      return this.data;
+
+    return new Promise((resolve, reject) => {
```
jlipps: so yeah, this would be rewritten as:

```js
return await B.promisify(openssl)('x509', {outform: 'der', in: this.pemFilename});
```

lib/certificate.js
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+}
+
+/**
+ * Interface for adding and removing records to TrustStore.sqlite3 databases that Keychains use
+ */
+class TrustStore {
+  constructor (iphoneSharedResourcesDir) {
+    this.sqliteDBPath = iphoneSharedResourcesDir + '/Library/Keychains/TrustStore.sqlite3';
```
jlipps: prefer template string over concatenation:

```js
this.sqliteDBPath = `${iphoneSharedResourcesDir}/Library/Keychains/TrustStore.sqlite3`;
```

(also, consider using `path.resolve` instead of this, since it will ensure absence/presence of trailing slash is not an issue)

(also, consider a shorter variable name than `iphoneSharedResourcesDir`, maybe `iphone` is redundant info at this point?)

lib/certificate.js
```diff
@@ -0,0 +1,162 @@
...
+    this.db = new sqlite3.Database(this.sqliteDBPath);
+  }
+
+  /**
+   * Add record to tsettings
+   */
+  async addRecord (sha1, tset, subj, data) {
+    return new Promise(async (resolve, reject) => {
```
jlipps: again, this could be rewritten as:

```js
let existingRecords = await this.getRecords(subj);
if (existingRecords.length > 0) {
  await B.promisify(this.db.run)(`UPDATE tsettings SET sha1=?, tset=?, data=? WHERE subj=?`, [sha1, tset, data, subj]);
}
...
```

(although, you might need to do something funky with bindings of `this.db.run` to make sure `run` is bound to `this.db`)

lib/utils.js
```diff
@@ -132,4 +133,23 @@ async function safeRimRaf (delPath, tryNum = 0) {
   }
 }
 
-export { killAllSimulators, endAllSimulatorDaemons, safeRimRaf, simExists };
+async function installSSLCert (pemText, udid) {
+  let tempFileName = `${__dirname}/temp-ssl-cert.pem`;
+  // TODO: Confirm if this is universally applicable to all versions of OSX
+  let pathToKeychain = `${process.env.HOME}/Library/Developer/CoreSimulator/Devices/${udid}/data/`;
```
jlipps: pretty sure you don't need to construct this dir here, I think we have methods already which get you a path like this from the udid?

test/unit/certificate-specs.js
```diff
@@ -0,0 +1,109 @@
...
+describe('when using TrustStore class', () => {
+
+  beforeEach(() => {
+    trustStore = new TrustStore(assetsDir);
+    testUUID = uuid.v4();
+  });
+
+  it('can add a record to the TrustStore tsettings', async () => {
+    let tsettings = await trustStore.getRecords(testUUID); 
+    chai.assert(tsettings.length===0);
```
jlipps: we use the `should` idiom:

```js
tsettings.length.should.equal(0);
```

or even more idiomatically:

```js
tsettings.should.have.length(0);
```

lib/certificate.js
dpgraham: I'll upgrade it to eslint-config-appium 0.1.0; it was at 0.0.6

dpgraham:
Thanks for the feedback @jlipps. I've implemented all of the changes you requested. I definitely like the .promisify syntax a lot better than what I was doing before.

lib/certificate.js
```diff
@@ -0,0 +1,142 @@
...
+    let subject = await this.getSubject(this.pemFilename);
+    let trustStore = new TrustStore(dir);
+    return trustStore.removeRecord(subject);
+  }
+
+  /**
+   * Translate PEM file to DER buffer
+   */
+  async getDerData () {
+    if (this.data) {
```
imurchie: As far as I can see in this class, `this.data` is _never_ set, so this check doesn't seem to be needed.

lib/certificate.js
```diff
@@ -0,0 +1,142 @@
...
+    this.pemFilename = pemFilename;
+  }
+
+  /**
+   * Add a certificate to the TrustStore
+   */
+  async add (dir) {
+    let data = await this.getDerData(this.pemFilename);
+    let subject = await this.getSubject(this.pemFilename);
+    let fingerprint = await this.getFingerPrint(this.data);
```
imurchie: Is `this.data` ever set?

lib/certificate.js
```diff
@@ -0,0 +1,142 @@
...
+    shasum.update(data);
+    this.fingerprint = shasum.digest();
+    return this.fingerprint;
+  }
+
+  /**
+   * Parse the subject from the der data
+   */
+  async getSubject () {
+    if (this.subject) {
```
imurchie: This `this.subject` is also never set.

lib/tail-until.js
```diff
@@ -17,9 +17,12 @@ async function tailUntil (filePath, until, timeout = 5000) {
 
   return new B(async (resolve, reject) => {
     let started = proc.start(startDetector);
+
+    /* eslint-disable */
```
imurchie: What is eslint complaining about here? It is usually better to disable it for a particular rule rather than in general.

dpgraham: The rule is ```promise/prefer-await-to-then```

I'll have it single out that particular rule

lib/certificate.js
dpgraham: This will be in the next push.

lib/certificate.js
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Translate PEM file to DER buffer
+   */
+  async getDerData () {
+    if (this.data) {
+      return this.data;
+    }
+
+    // Convert 'pem' file to 'der'
+    return openssl('x509', {
```
dpgraham: I'm having it set ```this.data = return openssl(...)```

The 'getDerData', 'getSubj', and 'getFingerprint' methods cache the results and then return them if they've been calculated already. I wanted lazy evaluation so that it doesn't do any unnecessary OpenSSL operations

imurchie: That seems like a good idea. At the moment I don't think it does that, though.

dpgraham:
Thanks @imurchie, I've made changes based on your review

imurchie:
@dpgraham Looks good to me! This was a lot of work, eh?

lib/certificate.js
```diff
@@ -0,0 +1,145 @@
...
+    this.db = new sqlite3.Database(this.sqliteDBPath);
+  }
+
+  /**
+   * Add record to tsettings
+   */
+  async addRecord (sha1, tset, subj, data) {
+    let existingRecords = await this.getRecords(subj);
+    if (existingRecords.length > 0) {
+      return this.db.runAsync(`UPDATE tsettings SET sha1=?, tset=?, data=? WHERE subj=?`, [sha1, tset, data, subj]);
```
jlipps: I'm concerned that we're not `awaiting` this. It feels like we're just getting lucky if this is completing before we need the result to be in the table. Might be a source of race condition bugs.

lib/tail-until.js
```diff
@@ -17,9 +17,12 @@ async function tailUntil (filePath, until, timeout = 5000) {
 
   return new B(async (resolve, reject) => {
     let started = proc.start(startDetector);
+
```
jlipps: I know you didn't write this particular section of code, but I'm pretty sure that `proc.start` includes the ability to send in a timeout, which would make the next few lines unnecessary...

test/assets/certificate-test-server/server.js
```diff
@@ -0,0 +1,100 @@
...
+    let server = https.createServer({key: keys.serviceKey, cert: keys.certificate}, function (req, res){
+      res.end('If you are seeing this the certificate has been installed');
+    }).listen(9758);
+
+    console.log('Make sure you have at least one IOS Simulator running'.yellow);
+    let devices = await getDevices();
+
+    // Get currently booted devices
+    let bootedDevices = [];
+    _.forEach(devices, (os) => {
```
jlipps: we prefer `for (let os of devices) {`

test/assets/certificate-test-server/server.js
```diff
@@ -0,0 +1,100 @@
...
+      os.forEach((device) => {
+        if(device.state === 'Booted'){
+          bootedDevices.push(device);
+        }
+      });
+    });
+
+    if(bootedDevices.length === 0) {
```
jlipps: `if (`

test/assets/certificate-test-server/server.js
```diff
@@ -0,0 +1,100 @@
...
+    let bootedDevice = bootedDevices[0];
+    let udid = bootedDevice.udid;
+    let deviceName = bootedDevice.name;
+
+    console.log('HTTPS server is running at localhost:9758 and has created a new certificate at "random-pem.pem"'.yellow);  
+    console.log(`Navigate to https://localhost:9758 in '${deviceName} Simulator'`.yellow);
+    console.log('DO NOT PUSH THE CONTINUE BUTTON. PUSH CANCEL.'.red);
+    
+    // Call this if the user answers 'No' to any prompts
+    async function done(){
```
jlipps: `done() {`

test/assets/certificate-test-server/server.js
```diff
@@ -0,0 +1,100 @@
...
+    console.log('Installing certificate'.yellow);
+    console.log(`Certificate installed to '${deviceName} ${udid}'. Navigate back to https://localhost:9758.`.yellow);
+
+    result = await inquirer.prompt([{
+      type: 'confirm',
+      name: 'confirmOpenedSite',
+      message: 'Now is https://localhost:9758 accessible?',
+    }]);
+
+    if(!result.confirmOpenedSite){
```
jlipps: `) {`

jlipps: `if (`

test/assets/certificate-test-server/server.js
```diff
@@ -0,0 +1,100 @@
...
+    console.log(`SSL cert removed.`.yellow);
+    console.log(`Close the simulator, re-open it and then navigate back to https://localhost:9758`.yellow);
+
+    result = await inquirer.prompt([{
+      type: 'confirm',
+      name: 'confirmUninstallCert',
+      message: `Is https://localhost:9758 unaccessible?`,
+    }]);
+
+    if(result.confirmUninstallCert){
```
jlipps: `if (` ... `) {`

lib/certificate.js
dpgraham: I'll add 'await' just to keep it consistent. I did actually have a race condition in previous tests because I wasn't returning it. I could see how there'd be a race condition if the addRecord is called with awaiting the result.

lib/tail-until.js
dpgraham: So should I delete that entire block that I have disabled?

dpgraham:
Changes applied again.

PR #83, opened 2016-12-09
Await stopping of proc

imurchie:


imurchie:
We shouldn't get any rejections, since we are checking the running state before doing the stop.

jlipps:
@imurchie what if `stop` itself throws for some other reason? i think you're right that we're fixing the problem we're seeing at the moment, but i don't think this PR is doing so in a way that prevents unhandled rejections in general. if we want to do that, i think we'd need something like:

```js
  }).finally(async () => {
    // no matter what, stop the tail process
    try {
      await proc.stop();
    } catch (ign) {
      // we can't really do anything if this fails for whatever reason,
      // so simply ignore it
    }
  });
}
```

Since we're catching any possible error, this is guaranteed to not have an unhandled rejection.

imurchie:
I like that. I'll update.

PR #85, opened 2016-12-15
Use different startup timeout for iOS 10+

imurchie:
Make the startup timeout dynamic, so that iOS 10+ sims can have a longer one. Why oh why is the simulator boot so slow?

This seems to fix, locally at least, the XCUITest issue in which `xcodebuild` errors out with `"Early unexpected exit, operation never finished bootstrapping - no restart will be attempted"`. This seems to happen when the device is not quite ready to get installed to.

### DIFF
--- a/lib/tail-until.js
+++ b/lib/tail-until.js
@@ -16,18 +16,16 @@ async function tailUntil (filePath, until, timeout = 5000) {
     return stdout.indexOf(until) > -1;
   };
 
-  return new B(async (resolve, reject) => {
+  return await new B((resolve, reject) => {
     let started = proc.start(startDetector);
 
     /* eslint-disable promise/prefer-await-to-then */
     let timedout = B.delay(timeout).then(() => {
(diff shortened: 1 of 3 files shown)


lib/tail-until.js
```diff
@@ -16,18 +16,24 @@ async function tailUntil (filePath, until, timeout = 5000) {
...
 
     /* eslint-disable promise/prefer-await-to-then */
     let timedout = B.delay(timeout).then(() => {
-      return reject(`tailing file ${filePath} failed after ${timeout}ms`);
+      return reject(new Error(`Tailing file ${filePath} failed after ${timeout}ms`));
     });
     /* eslint-enable */
 
-    await B.race([started, timedout]);
+    return (async () => {
```
jlipps: since we're already in a promise context here, I wonder if it's more readable to just do:

```js
B.race([started, timedout]).then(resolve).catch(reject);
```

imurchie: Sure!

PR #86, opened 2016-12-22
TrustStore creates keychains folder if none exists

dpgraham:
If there is no keychains folder at the TrustStore library, create one and create a new sqlite db with a tsettings table

**Reviewers: @jlipps @imurchie**

### DIFF
--- a/lib/certificate.js
+++ b/lib/certificate.js
@@ -133,7 +133,14 @@ class TrustStore {
 
     // If it doesn't have a tsettings table, create one
     await this.db.runAsync(`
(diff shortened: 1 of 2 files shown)


lib/certificate.js
```diff
@@ -110,34 +111,62 @@ class Certificate {
...
+    if (!(await fs.exists(keychainsPath))) {
+      await mkdirp(keychainsPath);
+    }
+
+    // Open sqlite database
+    this.db = new sqlite3.Database(path.resolve(keychainsPath, 'TrustStore.sqlite3'));
+
+    // If it doesn't have a tsettings table, create one
+    await this.db.runAsync(`
+      CREATE TABLE IF NOT EXISTS tsettings(sha1 BLOB NOT NULL DEFAULT '',subj BLOB NOT NULL DEFAULT '',tset BLOB,data BLOB,PRIMARY KEY(sha1));
```
jlipps: any way we can use multiline strings so this doesn't wrap weirdly?

PR #87, opened 2016-12-29
Lazy load the sqlite3 module

imurchie:
The `sqlite3` module seems to have problems that are hard to pin down, finding the native library it uses (see, for instance, https://github.com/appium/appium/issues/7624). Rather than globally load the module, wait until it is needed, so that the only people affected are the people that need it, and it effort can be made to fix it for them, rather than for lots of people for whom the whole thing is irrelevant.


jlipps:
👍

PR #89, opened 2017-02-02
Removed sqlite3 dependency

dpgraham:
-Uses teen_process exec to run sqlite3 from command line (sqlite3 comes packaged with OSX)
-Inserts sha1 and data in the hex format because they're both BLOBs
-Added a 'hasRecords' and 'getRecordCount' function that checks how many records are in tsettings and then parses the result
-Checks that 'sqlite3' is available on path before installing an SSL cert (just in case it was uninstalled or unavailable in a future release)
-Updated tests to use new code

Reviewers @jlipps @imurchie

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -146,7 +146,7 @@ async function installSSLCert (pemText, udid) {
 
   // Check that sqlite3 is installed on the path
   try {
-    await exec('which', ['sqlite3']);
+    await fs.which('sqlite3');
   } catch (e) {
     log.debug(`customSSLCert requires sqlite3 to be available on path`);
     log.errorAndThrow(`Command 'sqlite3' not found`);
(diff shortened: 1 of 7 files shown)


lib/utils.js
```diff
@@ -144,8 +144,16 @@ async function installSSLCert (pemText, udid) {
     log.errorAndThrow(`Command 'openssl' not found`);
   }
 
-  let tempFileName = path.resolve(__dirname, 'temp-ssl-cert.pem');
-  let pathToKeychain = new Simulator(udid).getDir();
+  // Check that sqlite3 is installed on the path
+  try {
+    await exec('which', ['sqlite3']);
```
imurchie: Does `await fs.which('sqlite3');` not work?

dpgraham: It does work. I'll update that for consistency.

imurchie:
I think something went wrong with rebasing, etc. Merge commits in there will make merging and squashing weird.

jlipps:
I squashed and merged using github's api! muhaha

jlipps:
published in `appium-simulator@1.11.1`

PR #91, opened 2017-02-03
Optimize the way we detect Simulator is booted

mykola-mokhnach:
Use process detection instead of looking for magic string in logs. This should make the detection more stable and shorten simulator startup duration.

### DIFF
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -58,9 +58,14 @@ class SimulatorXcode8 extends SimulatorXcode7 {
     const startupTimestamp = process.hrtime();
     try {
       await waitForCondition(async () => {
-        let {stdout} = await exec('bash', ['-c',
-          "ps axo command | grep Simulator | grep nsurlstoraged | grep -v bash | grep -v grep"]);
-        return stdout.trim().length > 0;
+        try {
+          let {stdout} = await exec('bash', ['-c',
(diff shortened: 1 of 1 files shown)


imurchie:
I like this! But on my machine (OSX 10.12.3) the command that gets run throws an error if nothing is found, so the contents of `waitForCondition` needed to be wrapped in a try/catch block or else the whole thing immediately exits with a failure.

```js
    try {
      await waitForCondition(async () => {
        try {
          let {stdout} = await exec('bash', ['-c',
            "ps axo command | grep Simulator | grep nsurlstoraged | grep -v bash | grep -v grep"]);
          return stdout.trim().length;
        } catch (err) {
          return false;
        }
      }, {waitMs: startupTimeout, intervalMs: 500});
    } catch (err) {
      log.errorAndThrow(`The Simulator is not booted after ${process.hrtime(startupTimestamp)[0]} seconds`);
    }
```

mykola-mokhnach:
Thanks @imurchie. Fixed.

mykola-mokhnach:
@jlipps It will be only possible to remove the redundant code if this detection method works for previous Xcode versions (6-7). Unfortunately, I don't have possibility to check it. That is why this only has been added for Xcode8.

jlipps:
Makes sense. Let's leave the others untouched for now. In 6 months we'll be removing support for XCode < 8 anyway!

imurchie:
Published in `appium-ios-simulator@1.12.0`.

PR #93, opened 2017-02-06
Make it possible to add onBootCompleted callback to Simulator

mykola-mokhnach:
IOS simulator allows to execute some xcrun terminal commands even if UI initialisation has not been 100% completed. This will allow us to shorten session init duration if we install/uninstall app under test in parallel while Simulator is still booting.


imurchie:
Looks good now!

imurchie:
Published in `appium-ios-simulator@1.13.0`.

jlipps:
Yes, great idea, thanks @mykola-mokhnach!

PR #94, opened 2017-02-06
Speed up Safari startup

mykola-mokhnach:
Because we can do it (and get a t-shirt  :-P )

### DIFF
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -73,7 +73,7 @@ class SimulatorXcode8 extends SimulatorXcode7 {
     } catch (err) {
       log.errorAndThrow(`Safari cannot be started after ${process.hrtime(launchTimestamp)[0]} seconds`);
     }
-    log.debug(`Safari has successfully started and opened ${url} in ${process.hrtime(launchTimestamp)[0]} seconds`);
+    log.debug(`Safari has successfully opened ${url} in ${process.hrtime(launchTimestamp)[0]} seconds`);
   }
 
 }
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -40,7 +40,7 @@ class SimulatorXcode8 extends SimulatorXcode7 {
           if (stdout.trim().length > 0) {
             // 'springboard' process should be the last one to start after boot
             // 'simctl launch' will block until this process is running
-            await exec('bash', ['-c', `xcrun simctl launch ${this.udid} com.apple.springboard`]);
+            await exec('xcrun', ['simctl', 'launch', this.udid, 'com.apple.springboard']);
             return true;
           }
         } catch (e) {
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
The current implementation works properly and fast only in case Safari application is not opened yet in Sim. In case we want to use this method for changing currently opened url in browser then let me know and I'll modify it.

lib/simulator-xcode-8.js
```diff
@@ -53,6 +53,29 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
+      throw new Error(`Tried to open ${url}, but Simulator is not in Booted state`);
+    }
+    const launchTimestamp = process.hrtime();
+    try {
+      await waitForCondition(async () => {
+        try {
+          await exec('bash', ['-c', `xcrun simctl launch ${this.udid} com.apple.mobilesafari ${url}`]);
+          return true;
+        } catch (e) {
+          log.error(e);
```
imurchie: I think printing out the whole error (with stack trace and all) at this point is a recipe for people being confused and thinking something is wrong. Perhaps a waiting message (rate limited, if this takes a while) would be better.

lib/simulator-xcode-8.js
```diff
@@ -53,6 +53,29 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
+    const SAFARI_STARTUP_TIMEOUT = 15 * 1000;
+
+    if (!await this.isRunning()) {
+      throw new Error(`Tried to open ${url}, but Simulator is not in Booted state`);
+    }
+    const launchTimestamp = process.hrtime();
+    try {
+      await waitForCondition(async () => {
+        try {
+          await exec('bash', ['-c', `xcrun simctl launch ${this.udid} com.apple.mobilesafari ${url}`]);
```
imurchie: What is the reason for running this through `bash` instead of just running `xcrun` (like `await exec('xcrun', ['simctl', 'launch', this.udid, 'com.apple.mobilesafari', url]);`)?

lib/simulator-xcode-8.js
```diff
@@ -53,6 +53,29 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
+          return true;
+        } catch (e) {
+          log.error(e);
+        }
+        return false;
+      }, {waitMs: SAFARI_STARTUP_TIMEOUT, intervalMs: 500});
+    } catch (err) {
+      log.errorAndThrow(`Safari cannot be started after ${process.hrtime(launchTimestamp)[0]} seconds`);
```
imurchie: I think the error message should probably get into here somewhere, so it is not entirely opaque.

lib/simulator-xcode-8.js
mykola-mokhnach: I just copied the call from some another existing method and that is why though, that called it the other way makes issues on Travis, etc. But if you say it is no problem to call xcrun directly...

lib/simulator-xcode-8.js
mykola-mokhnach: OK, agree

lib/simulator-xcode-8.js
mykola-mokhnach: you mean the **err** ?

lib/simulator-xcode-8.js
imurchie: I think it is only necessary if you are doing things like piping from on process to another (which `exec` cannot handle on its own, so `bash` is needed). Otherwise it is forking `bash` in order for it to fork `xcrun`, which is heavier than necessary, I think.

lib/simulator-xcode-8.js
imurchie: Yes. Some indication of what the error is. Most of the time it will be that the browser didn't open and timed out. But there have been cases where this sort of pattern has hidden some configurational error or another, and it would _never_ work but was not apparent.

imurchie: Something like
```js
log.errorAndThrow(`Safari cannot be started after ${process.hrtime(launchTimestamp)[0]} seconds: ${err.message}`);
```

mykola-mokhnach: Won't this error be always the same (waitForCondition timeout)? Or it can also provide internal backtrace?

imurchie: Hm... yes, that's true.

PR #96, opened 2017-02-08
Add killall signal in Sierra

imurchie:
On 10.12 the killing of sims is often flakey. sending the killall command is much more reliable.


imurchie:
Well, the whole `killAllSimulators` thing is a problem for that. This just makes it more likely that it will do what it says. If we are exploring the multiple sim thing, this remains the one place where everything is killed.

jlipps:
👍  makes the sense

PR #97, opened 2017-02-09
Add simulator argument to disconnect hardware keyboard by default

mykola-mokhnach:
This helps to workaround known XCTest issue described in https://github.com/facebook/WebDriverAgent/issues/453

I think it's safe to have this option included by default, since it does not affect anything while interacting with the Simulator via Appium. Although, we could also add a separate capability to have it enabled or disabled, if you think this can be somehow useful.


imurchie:
I'm good with this as the only way. If someone eventually finds it onerous, we can make it configurable.

PR #98, opened 2017-02-09
Do not try to start the simulator if there are no folders

mykola-mokhnach:
I suppose there is no need to start and stop Simulator if there is no directory for the given app. Returning **undefined** means the app has no data to cleanup, which is OK for us. And starting the simulator and then quitting it is a very time-expensive operation.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -100,16 +100,10 @@ class SimulatorXcode6 extends EventEmitter {
    * or, for iOS 7.1, the app name without `.app` (e.g., MobileSafari)
    */
   async getAppDir (id, subDir = 'Data') {
-    if (await this.isFresh()) {
-      log.info('Attempted to get an app path from a fresh simulator ' +
-               'quickly launching the sim to populate its directories');
-      await this.launchAndQuit();
-    }
-
-    if (!this.appDataBundlePaths[subDir]) {
+    this.appDataBundlePaths[subDir] = this.appDataBundlePaths[subDir] || {};
+    if (_.isEmpty(this.appDataBundlePaths[subDir]) && !await this.isFresh()) {
       this.appDataBundlePaths[subDir] = await this.buildBundlePathMap(subDir);
     }
-
     return this.appDataBundlePaths[subDir][id];
   }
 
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-6.js
```diff
@@ -100,16 +100,10 @@ class SimulatorXcode6 extends EventEmitter {
    * or, for iOS 7.1, the app name without `.app` (e.g., MobileSafari)
    */
   async getAppDir (id, subDir = 'Data') {
-    if (await this.isFresh()) {
-      log.info('Attempted to get an app path from a fresh simulator ' +
-               'quickly launching the sim to populate its directories');
-      await this.launchAndQuit();
```
jlipps: I don't understand. In order to make changes to app directory files (e.g., to set certain flags for the purpose of say the `locationServicesEnabled` capability), we have to ensure that the app has previously been launched. In our previous experimentation, setting the flag in a running sim did not take effect; we had to launch the app, modify the plist, then shut down the sim in order for it to reliably work. It looks like you're removing this behavior?

mykola-mokhnach: Simulator restart is just a side effect for the **getAppDir** method.  If we really need to stop and start Simulator for such cases then it should happen in the corresponding methods where these properties are being set.

imurchie: When updating settings, etc., we manually use `isFresh` and `launchAndQuit`. See, for instance, https://github.com/appium/appium-ios-driver/blob/5393545c51f983873e34b5919e78ab1ad25cfbcc/lib/settings.js#L39

mykola-mokhnach: so then there should be no problems with this change, rigth @imurchie ?

imurchie: Well... the xcuitest test all pass with it. I'm going to try the ios driver tests.

jlipps: excellent point. if we intend to launch and quit for a particular purpose, we should do it in the place where that purpose is defined in code, not as a hidden side effect here. 👍

imurchie:
Looks good.

imurchie:
Published in `appium-ios-simulator@1.16.0`.

PR #102, opened 2017-02-27
Tune simulator shutdown procedure

mykola-mokhnach:
This PR addresses complicated simulator cases when it cannot be terminated easily, for example https://github.com/appium/appium/issues/7904

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -49,12 +49,13 @@ async function killAllSimulators (timeout = DEFAULT_SIM_SHUTDOWN_TIMEOUT) {
     log.warn(`pgrep error ${e.code} while detecting whether ${appName} is running. Trying to kill anyway.`);
   }
 
-  const rc = await killApp(appName, false);
-  if ([0, 1].indexOf(rc) !== -1) {
+  let rc = await killApp(appName, false);
+  if (rc === 1) {
     log.debug(`${appName} is not running. Continuing...`);
     return;
+  } else if (rc > 1) {
+    log.warn(`pkill error ${rc} while trying to gracefully kill ${appName}`);
   }
-  log.warn(`pkill error ${rc} while trying to gracefully kill ${appName}`);
 
   // wait for all the devices to be shutdown before Continuing
   // but only print out the failed ones when they are actually fully failed
@@ -86,7 +87,10 @@ async function killAllSimulators (timeout = DEFAULT_SIM_SHUTDOWN_TIMEOUT) {
         log.warn(`    ${device}`);
       }
       log.warn('Sending force kill signal to shutdown them all');
-      await killApp(appName, true);
+      rc = await killApp(appName, true);
+      if ([0, 1].indexOf(rc) === -1) {
+        log.errorAndThrow(`Cannot forcefully terminate ${appName}. pkill error code: ${rc}`);
+      }
       return;
     }
     throw err;
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -49,12 +49,16 @@ async function killAllSimulators (timeout = DEFAULT_SIM_SHUTDOWN_TIMEOUT) {
     log.warn(`pgrep error ${e.code} while detecting whether ${appName} is running. Trying to kill anyway.`);
   }
 
-  let rc = await killApp(appName, false);
-  if (rc === 1) {
-    log.debug(`${appName} is not running. Continuing...`);
+  try {
+    await exec('xcrun', ['simctl', 'shutdown', 'booted'], {timeout: timeout});
     return;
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -4,60 +4,62 @@ import { exec } from 'teen_process';
...
+async function forceKillAll (appName) {
   try {
-    await exec('killall', [appName]);
-  } catch (err) {
-    // if no process exists we'll get an error that we expect.
-    // otherwise throw it again
-    if (err.stderr.indexOf(`No matching processes belonging to you were found`) === -1) {
-      throw err;
+    await exec('pkill', ['-9', '-x', appName]);
+  } catch (e) {
```
mykola-mokhnach: @imurchie I'm not 100% sure here about teen_process lib functionality. Does it throw an exception if the exit code differs from zero?

imurchie: It should, yes.

https://github.com/appium/node-teen_process/blob/master/lib/teen_process.js#L67-L73

lib/utils.js
```diff
@@ -4,60 +4,62 @@ import { exec } from 'teen_process';
...
+    if (e.code === 1) {
+      log.debug(`${appName} is not running. Continuing...`);
+      return;
     }
+    log.error(`pkill error ${e.code} while trying to forcefully kill ${appName}`);
```
jlipps: `log.errorAndThrow` is made for such a case

lib/utils.js
```diff
@@ -4,60 +4,62 @@ import { exec } from 'teen_process';
...
-    let macVersion = await system.macOsxVersion();
-    if (macVersion === '10.12') {
-      await sendKillAll(appName);
+    await exec('pgrep', ['-x', appName]);
+  } catch (e) {
+    if (e.code === 1) {
+      log.debug(`${appName} is not running. Continuing...`);
+      return;
     }
+    log.error(`pgrep error ${e.code} while detecting whether ${appName} is running. Trying to kill anyway...`);
```
jlipps: maybe this should be `warn` level if we're continuing anyway?

lib/utils.js
```diff
@@ -4,60 +4,62 @@ import { exec } from 'teen_process';
...
+  }
+
+  try {
```
jlipps: this block of code is basically the same as what's factored out in a method above, maybe we can consolidate these into one method which also has a `-9`-type parameter?

lib/utils.js
mykola-mokhnach: fixed

lib/utils.js
mykola-mokhnach: changed to warn

lib/utils.js
mykola-mokhnach: extracted to common method

lib/utils.js
```diff
@@ -18,16 +18,17 @@ const DEFAULT_SIM_SHUTDOWN_TIMEOUT = 10000;
...
+    return 0;
   } catch (e) {
-    if (e.code === 1) {
-      log.debug(`${appName} is not running. Continuing...`);
-      return;
-    }
-    log.error(`pkill error ${e.code} while trying to forcefully kill ${appName}`);
-    throw e;
+    if (e.code) return e.code;
+    return e;
```
jlipps: I never see us doing anything with a returned `Error` from this function; do we not want to `throw` if we can't find a code?

mykola-mokhnach: Probably we do want %)

lib/utils.js
```diff
@@ -4,60 +4,61 @@ import { exec } from 'teen_process';
...
-    // if no process exists we'll get an error that we expect.
-    // otherwise throw it again
-    if (err.stderr.indexOf(`No matching processes belonging to you were found`) === -1) {
-      throw err;
-    }
+    await exec('pkill', args);
+    return 0;
+  } catch (e) {
+    if (!_.isUndefined(e.code)) return e.code;
+    log.error(`Got unexpected error while trying to kill ${appName}`);
```
jlipps: not to nitpick, but it would be nice to know what the error is in this case:

```js
log.errorAndThrow(`Unexpected error while trying to kill ${appName}: ${e.message}`);
```

mykola-mokhnach: I expect the following throw to show the full stacktrace

mykola-mokhnach:
@jlipps @imurchie I did some changes to make Travis happy. Can you please take another look?

imurchie:
Looks good to me. All tests pass on xcuitest and ios drivers.

imurchie:
Published in `appium-ios-simulator@1.19.0`.

PR #103, opened 2017-03-07
enableCalendarAccess API

dpgraham:
* Moved execSQLiteQuery into a separate helper class and updated certificate.js to use that helper method
* Added calendar class that has methods for accessing TCC.db and adding an entry that approves calendar access on IOS Simulator
* Wrote unit tests that confirms that the entries are being created successfully and varying circumstances
* No functional tests yet. That will come in appium-xcuitest-driver. But I have confirmed that this works successfully by manually testing it on a small sample app.

(please review @jlipps @imurchie)


lib/calendar.js
```diff
@@ -0,0 +1,62 @@
+import Simulator from './simulator-xcode-6';
+import path from 'path';
+import { fs, mkdirp } from 'appium-support';
+import { execSQLiteQuery } from './utils';
+
+class Calendar {
```
jlipps: i've seen this pattern and it's not a bad one, but we've typically stayed away from classes unless there's a real good reason to maintain state, and this class doesn't have any state at all. i'm not sure we need it just as a namespace. maybe simple functions would do just as well?

dpgraham:
I updated that. I also added a 'disableCalendarAccess' for the use case where the calendar is explicitly disabled. Also, the syntax is a little weird with the `let Calendar;` at the beginning and then exporting `Calendar = {...}`, but I needed to do that in order to get sinon working because I needed to stub the methods and you can't stub if you're just exporting standard es6 methods.

lib/certificate.js
```diff
@@ -144,17 +144,17 @@ class TrustStore {
    */
   async addRecord (sha1, tset, subj, data) {
     if (await this.hasRecords(subj)) {
-      return await this.execSQLiteQuery(`UPDATE tsettings SET sha1=x'?', tset='?', data=x'?' WHERE subj='?'`, sha1, tset, data, subj);
+      return await execSQLiteQuery(await this.getDB(), `UPDATE tsettings SET sha1=x'?', tset='?', data=x'?' WHERE subj='?'`, sha1, tset, data, subj);
```
imurchie: It would be cleaner to pull the `await this.getDB()` out to before the conditional, since both cases use it.

lib/calendar.js
```diff
@@ -0,0 +1,72 @@
...
+  let count = await execSQLiteQuery(db, `SELECT count(*) FROM access WHERE client='?';`, bundleID);
+  return parseInt(count.stdout.split('=')[1], 10);
+}
+
+/**
+ * Allow calendar access for UDID
+ * @param {string} udid 
+ * @param {string} bundleID 
+ */
+async function enableCalendarAccess (udid, bundleID) {
```
imurchie: Since `enableCalendarAccess` and `disableCalendarAccess` are both taking the `udid` of a simulator, and do their business by creating that simulator object and getting the path from it, it seems to me that these functions ought to be on the simulator object itself. The driver that is trying to enable or disable the calendar has access to the simulator object, so the [calling code](https://github.com/appium/appium-xcuitest-driver/blob/9e6b357a2e49ac2276b71e4c4a07e7e0411c7053/lib/driver.js#L329-L338) goes from
```js
    if (!this.isRealDevice()) {
      // Enable calendar access if calendarAccessAuthorized flag is set
      // Disable calendar access if it is explicitly set to false
      if (this.opts.calendarAccessAuthorized) {
        enableCalendarAccess(this.opts.udid, this.opts.bundleId);
      } else if (this.opts.calendarAccessAuthorized === false) {
        disableCalendarAccess(this.opts.udid, this.opts.bundleId);
      }
    }
```
to
```js
    if (!this.isRealDevice()) {
      // Enable calendar access if calendarAccessAuthorized flag is set
      // Disable calendar access if it is explicitly set to false
      if (this.opts.calendarAccessAuthorized) {
        this.opts.device.enableCalendarAccess(this.opts.bundleId);
      } else if (this.opts.calendarAccessAuthorized === false) {
        this.opts.device.disableCalendarAccess(this.opts.bundleId);
      }
    }
```

dpgraham: Yeah that's much better. I'll change it to that.

jlipps: yeah, I agree with isaac's point

dpgraham:
I've made the changes that @imurchie has suggested. 

I refactored 'calendar.js' as it's own standalone class and then the simulator instantiates that class, passes it a sharedResourcesDir & bundleID, and calls enableCalendarAccess/disableCalendarAccess on it.

Also added a 'hasCalendarAccess' method to check if the db has that entry and allowed = 1

lib/simulator-xcode-6.js
```diff
@@ -571,6 +572,18 @@ class SimulatorXcode6 extends EventEmitter {
     }
   }
 
+  async enableCalendarAccess (bundleID) {
+    await (new Calendar(this.getDir(), bundleID)).enableCalendarAccess();
```
jlipps: this still feels a bit wonky to me. I think it's because we're creating Calendar objects every time, which nullifies any value in keeping state on an object. what if instead we did something like:

```js
this.calendar = new Calendar(this.getDir());
```

in the constructor. Then here we could do:

```js
await this.calendar.enableCalendarAccess(bundleID);
```

to me this makes more sense because bundleID isn't really an essential property of the Calendar class as you've defined it; I think it makes more sense to allow a sim object to have a cal object.

either that or we keep a hash of cal objects by bundle id on the sim object.

dpgraham: Done. I pass the bundleID every time now.

lib/calendar.js
```diff
@@ -0,0 +1,71 @@
...
+    return this.db;
+  }
+
+  async getCalendarRowCount (bundleID) {
+    let db = await this.getDB();
+    let count = await execSQLiteQuery(db, `SELECT count(*) FROM access WHERE client='?' AND service='kTCCServiceCalendar';`, bundleID);
+    return parseInt(count.stdout.split('=')[1], 10);
+  } 
+
+  async  hasCalendarAccess (bundleID) {
```
jlipps: looks like an extra space here?

lib/simulator-xcode-6.js
```diff
@@ -571,6 +574,18 @@ class SimulatorXcode6 extends EventEmitter {
...
+  async enableCalendarAccess (bundleID) {
+    await this.calendar.enableCalendarAccess(bundleID);
+  }
+
+  async disableCalendarAccess (bundleID) {
+    await this.calendar.disableCalendarAccess(bundleID);
+  }
+
+  async hasCalendarAccess (bundleID) {
+    return this.calendar.hasCalendarAccess(bundleID);
```
jlipps: nitpick:

even though these two lines are functionally equivalent:

```js
return this.calendar.hasCalendarAccess(bundleId);
return await this.calendar.hasCalendarAccess(bundleId);
```

we prefer the latter in our code because it signals to the reader that we are actually calling an async function, which you might not know otherwise, and be tempted to do something like:

```js
this.calendar.hasCalendarAccess(bundleID);
```

which is NOT the same as

```js
await this.calendar.hasCalendarAccess(bundleID);
```

lib/calendar.js
dpgraham: Done

dpgraham:
Squashed and merged

dpgraham:
Published in 1.20.0

PR #106, opened 2017-03-16
Added enrollTouchID() method to Simulator

dpgraham:
* Wrote bindTouchIDKey() method which adds keyboard shortcut to Touch ID Enrollment menu items
* binds key on startup, unbinds key when it's killed
* Wrote enrollTouchID() method that calls Applescript to toggle the above keyboard shortcut
* Wrote unit test that checks that the call is not rejected (can't test the functionality in Simulator, those tests will be in XCUITest Driver)

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -5,7 +5,8 @@ import log from './logger';
 import { fs } from 'appium-support';
 import B from 'bluebird';
 import _ from 'lodash';
-import { killAllSimulators, safeRimRaf, bindTouchIDKey } from './utils.js';
+import { killAllSimulators, safeRimRaf } from './utils.js';
+import { setTouchEnrollKey } from './touch-enroll.js';
 import { asyncmap, waitForCondition, retryInterval, retry } from 'asyncbox';
 import * as settings from './settings';
 import { exec } from 'teen_process';
@@ -290,7 +291,10 @@ class SimulatorXcode6 extends EventEmitter {
     if (!this.connectHardwareKeyboard) {
       args.push('-ConnectHardwareKeyboard', '0');
     }
-    await bindTouchIDKey();
+
+    // Do the 'Touch ID Enroll' key bindings before the Simulator starts
+    await setTouchEnrollKey();
+
     log.info(`Starting simulator with command: open ${args.join(' ')}`);
     let startTime = Date.now();
     await exec('open', args, {timeout: OPEN_TIMEOUT});
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -5,7 +5,7 @@ import log from './logger';
 import { fs } from 'appium-support';
 import B from 'bluebird';
 import _ from 'lodash';
-import { killAllSimulators, safeRimRaf } from './utils.js';
+import { killAllSimulators, safeRimRaf, bindTouchIDKey } from './utils.js';
 import { asyncmap, waitForCondition, retryInterval, retry } from 'asyncbox';
 import * as settings from './settings';
 import { exec } from 'teen_process';
@@ -290,6 +290,7 @@ class SimulatorXcode6 extends EventEmitter {
     if (!this.connectHardwareKeyboard) {
       args.push('-ConnectHardwareKeyboard', '0');
     }
+    await bindTouchIDKey();
     log.info(`Starting simulator with command: open ${args.join(' ')}`);
     let startTime = Date.now();
     await exec('open', args, {timeout: OPEN_TIMEOUT});
@@ -592,6 +593,15 @@ class SimulatorXcode6 extends EventEmitter {
 
   async hasCalendarAccess (bundleID) {
     return await this.calendar.hasCalendarAccess(bundleID);
+  }
+
+  async enrollTouchID () {
+    await exec('osascript', ['-e', `
+      activate application "Simulator"
+      tell application "System Events"
+        key code 17 using {control down, shift down, option down, command down}
+      end tell
+    `]);
   }
 
   static async _getDeviceStringPlatformVersion (platformVersion) {
--- a/lib/touch-enroll.js
+++ b/lib/touch-enroll.js
@@ -1,8 +1,4 @@
-import { exec as nodeExec } from 'child_process';
-import B from 'bluebird';
-
-// Had problems with teen_process, resorted to this
-const exec = B.promisify(nodeExec);
+import { exec } from 'teen_process';
 
 let touchEnrollMenuKeys = ['Toggle Enrolled State', 'Touch ID Enrolled'];
 let touchEnrollBackups;
@@ -39,11 +35,11 @@ async function getUserDefault (domain, key) {
   let res;
   try {
     // If it doesn't find anything for this domain, it will throw an error so return undefined
-    res = await exec(`"defaults" read "Apple Global Domain" ${domain}`);
+    res = await exec('defaults', ['read', 'Apple Global Domain', domain]);
   } catch (stderr) {
     return;
   }
-  let stdout = res[0];
+  let stdout = res.stdout;
 
   // Parse the result into a Javascript array
   let nsUserKeyArr = stdout.trim()
@@ -74,8 +70,7 @@ async function getUserDefault (domain, key) {
  * @param {*} value 
  */
 async function setUserDefault (domain, key, value) {
-  let quotedValue = `'${value}'`;
-  await exec(`defaults write "Apple Global Domain" ${domain} -dict-add "${key}" ${typeof(value) === 'undefined' ? 'nil' : quotedValue}`);
+  await exec('defaults', ['write', 'Apple Global Domain', domain, '-dict-add', key, typeof(value) === 'undefined' ? 'nil' : value]);
 }
 
 async function backupTouchEnrollShortcuts () {
--- a/lib/touch-enroll.js
+++ b/lib/touch-enroll.js
@@ -1,0 +1,95 @@
+import { exec as nodeExec } from 'child_process';
+import B from 'bluebird';
+
+// Had problems with teen_process, resorted to this
+const exec = B.promisify(nodeExec);
+
+let touchEnrollMenuKeys = ['Toggle Enrolled State', 'Touch ID Enrolled'];
+let touchEnrollBackups;
+const NS_USER_KEY_EQUIVALENTS = 'NSUserKeyEquivalents';
+const TOUCH_ENROLL_KEY_CODE = '@~$^t';
+
+async function setTouchEnrollKey () {
+  await backupTouchEnrollShortcuts();
+  for (let key of touchEnrollMenuKeys) {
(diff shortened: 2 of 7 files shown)


imurchie:
This looks good. My only concern is messing with the defaults all the time. What if the user already has something set for that simulator, and after running Appium it is suddenly in a different state?

dpgraham:
Is there a way to backup the current default and then restore it?

imurchie:
Yes. If you do not have anything set,
```
$ defaults read "Apple Global Domain" NSUserKeyEquivalents "Toggle Enrolled State"
2017-03-16 14:11:03.306 defaults[44520:902404]
The domain/default pair of (kCFPreferencesAnyApplication, NSUserKeyEquivalents) does not exist
```

After running `defaults write "Apple Global Domain" NSUserKeyEquivalents -dict-add "Toggle Enrolled State" '@~\\$^t'`
```
$ defaults read "Apple Global Domain" NSUserKeyEquivalents "Toggle Enrolled State"
{
    "Toggle Enrolled State" = "@~\\\\$^t";
}

**Comment by dpgraham** [comment]
I'll update it so that we save the key at the startup and then restore the saved key when it closes down.

**Comment by dpgraham** [comment]
Changes made.

One comment is that I had to use child_process instead of teen_process.

The calls to `defaults` weren't taking effect when using the array syntax (`exec('defaults', [...]')`) but was taking effect when using the child_process string syntax (`exec('"defaults" ...')`)

File: lib/utils.js [hunk 1]
```diff
@@ -223,11 +226,4 @@ async function execSQLiteQuery (db, query, ...queryParams) {
   return await exec('sqlite3', ['-line', db, formattedQuery.join('')]);
 }
 
-async function bindTouchIDKey (bind = true) {
-  // JSHint ignore is to ignore the bad/unnecessary escaping rule
-  ['Toggle Enrolled State', 'Touch ID Enrolled'].forEach(async (key) => 
-    await exec('defaults', ['write', '"Apple Global Domain"', 'NSUserKeyEquivalents', '-dict-add', `"${key}"`, bind ? '"@~\\$^t"': 'nil'])
```
jlipps: I'm guessing the reason teen_process's exec wasn't working is that you have too many quotes in here. teen_process will handle quoting. what if this line were just:

```js
await exec('defaults', ['write', 'Apple Global Domain', 'NSUserKeyEquivalents', '-dict-add', `${key}`, bind ? '@~$^t': 'nil']));
```

(teen_process uses a shlex library which should take care of shell escaping and quoting)

lib/simulator-xcode-6.js
```diff
@@ -290,6 +291,10 @@ class SimulatorXcode6 extends EventEmitter {
     if (!this.connectHardwareKeyboard) {
       args.push('-ConnectHardwareKeyboard', '0');
     }
+
+    // Do the 'Touch ID Enroll' key bindings before the Simulator starts
+    await setTouchEnrollKey();
```
jlipps: also remember that this whole behavior will be behind a desired cap, so we shouldn't do it every time. we should only do it if a certain param is passed in to the ios sim constructor.

lib/touch-enroll.js
```diff
@@ -0,0 +1,95 @@
...
+ * @param {*} domain {string}
+ * @param {*} key {string|number|boolean}
+ */
+async function getUserDefault (domain, key) {
+  let res;
+  try {
+    // If it doesn't find anything for this domain, it will throw an error so return undefined
+    res = await exec(`"defaults" read "Apple Global Domain" ${domain}`);
+  } catch (stderr) {
+    return;
```
jlipps: should probably at least log here, something like "Did not find any user defaults for touch enroll keys"

lib/touch-enroll.js
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Get MacOS User Defaults by domain and key (for reference: `man defaults`)
+ * @param {*} domain {string}
+ * @param {*} key {string|number|boolean}
+ */
+async function getUserDefault (domain, key) {
+  let res;
+  try {
+    // If it doesn't find anything for this domain, it will throw an error so return undefined
+    res = await exec(`"defaults" read "Apple Global Domain" ${domain}`);
```
jlipps: can't think of why you'd need to quote "defaults", is there a reason?

lib/touch-enroll.js
```diff
@@ -0,0 +1,95 @@
+import { exec as nodeExec } from 'child_process';
+import B from 'bluebird';
+
+// Had problems with teen_process, resorted to this
+const exec = B.promisify(nodeExec);
+
+let touchEnrollMenuKeys = ['Toggle Enrolled State', 'Touch ID Enrolled'];
+let touchEnrollBackups;
```
jlipps: this is global to the file; what if multiple tests are running at once? should this be a class if we need to maintain state? or we could maintain state on the ios sim object?

imurchie:
Strange behavior. With `teen_process` getting the defaults works fine. Setting them throws an error, says no defaults were changed, and changes them anyway.

```js
    let domain = 'FakeDomain';
    try {
      let {stdout} = await exec('defaults', ['read', 'Apple Global Domain', domain]);
      console.log(stdout);
    } catch (err) {
      console.log(err.stderr);
    }

    try {
      let key = 'yugerkey';
      let quotedValue = '"something else great"';
      await exec('defaults', ['write', 'Apple Global Domain', domain, '-dict-add', key, quotedValue]);
      let {stdout} = await exec('defaults', ['write', 'Apple Global Domain', domain]);

      ({stdout} = await exec('defaults', ['read', 'Apple Global Domain', domain]));
      console.log(stdout);
    } catch (err) {
      console.log(err);
    }
```

Before running:
```
{
    greatkey = "something awesome";
    yugekey = "something great";
}
```

Output from running JS code:
```
$ defaults read "Apple Global Domain" FakeDomain
{
    greatkey = "something awesome";
    yugekey = "something great";
}

{ [Error: Command 'defaults write 'Apple Global Domain' FakeDomain' exited with code 1]
  stdout: '',
  stderr: '2017-03-17 17:13:32.942 defaults[31295:4980480] \nRep argument is not a dictionary\nDefaults have not been changed.\n',
  code: 1 }
```

After running:
```
$ defaults read "Apple Global Domain" FakeDomain
{
    greatkey = "something awesome";
    yugekey = "something great";
    yugerkey = "something else great";
}
```

lib/utils.js
dpgraham: Yeah that makes a lot of sense actually.

test/unit/touch-enroll-specs.js
```diff
@@ -0,0 +1,169 @@
...
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('touch-enroll.js', () => {
+
+  describe('getUserDefaults(), setUserDefaults()', () => {
+    let domain = 'fakeDomain';
+
+  // Nuke the contents of the domain so it's fresh every test case
+    let clearFakeDomain = async function () {
```
imurchie: This should really be using stubs so nothing on the system is changed.

lib/touch-enroll.js
dpgraham: Yeah that's part of the string syntax for exec (at least the documentation I looked at). I'll change this back to array syntax anyways so it won't be an issue.

imurchie: Also, what exactly is `exec` here. You've imported `child_process` `exec` as `nodeExec`.

test/unit/touch-enroll-specs.js
dpgraham: The defaults are applied to a non-existent domain called 'fakeDomain' and then when the tests are completed, the 'fakeDomain' is deleted so that the system is returned to it's original state. 

I don't know if there's still a risk associated with that or not, but if there is I suppose I could switch over to using stubs. Were you thinking I should stub `exec` to control the command line output.

imurchie: I don't know that there is any risk at all. I just don't think unit tests should be doing _anything_ real. The inputs and the outputs should be known. What if the test fails somewhere before resetting this fake domain?

lib/touch-enroll.js
dpgraham: I think this is a special case where we need global state. The `touchEnrollBackups` should only be set once, regardless of how many touchs `enrollTouchId` is called and then it should be restored when all of the simulators are killed.

jlipps: ok, that makes sense

lib/touch-enroll.js
dpgraham: If it throws an error, it doesn't mean that something went wrong, it means that the user hasn't set any defaults for that domain and thus the call to `exec('defaults read ...)` returned a status code !== 0. I think in this case returning `undefined` is acceptable and we don't need to log anything.

lib/simulator-xcode-6.js
dpgraham: Is it a desired capability or an endpoint or both? It seems like those two ought to be mutually exclusive (ie: either setting a desired capability `enrollTouchId=true` enables touch ID or calling the endpoint `/wd/hub/touch_id_enroll` toggles touchID enrollment).

If we use a desired capability, do we even need to be able to toggle touchID enrollment? Are there possible e2e test cases that would involve a device being enrolled in touchID and then unenrolled?

dpgraham:
@imurchie The call to `defaults write` is missing -dict-add <key> <value>

```javascript
let {stdout} = await exec('defaults', ['write', 'Apple Global Domain', domain]);
```

lib/simulator-xcode-6.js
dpgraham: EDIT: On further reading, I can see that there actually would be test cases where you need to test cases where touchID is enrolled and cases where touchID is not unenrolled. So it does make sense to have both.

imurchie:
Oh shoot, yeah. I had copied the one and forgotten to delete it. Anyway, suffice it to say that `teen_process` works.

lib/simulator-xcode-6.js
jlipps: Yep, you got it. The reason for the cap is to avoid this system tomfoolery for those who don't need the functionality, just to be safe.

lib/touch-enroll.js
dpgraham: `exec` is the promisified form of `child_process.exec`. This will be changed back to teen_process though.

test/unit/touch-enroll-specs.js
dpgraham: What if I broke up the tests into a `touch-enroll-specs.js` which uses mocks/stubs (the way unit tests are supposed to) and then I moved the system altering tests to `touch-enroll-e2e.js`? I just want to have tests in place that verify the correctness of the calls to `exec('default...')`

dpgraham:
Revisions pushed

* I added `teen_process` back and it works fine
* I moved the tests over to e2e and then added a few basic unit tests

imurchie:
Published in `appium-ios-simulator@1.23.0`.

PR #107, opened 2017-03-21
Allow touchEnroll only if variable set to true

dpgraham:
* Added 'allowTouchEnroll' as a param to `run` function
* If true, setTouchEnrollKeys is called
* Is not called if `allowTouchEnroll` is false
* Added test to check that enrollKeys are only backed up when `allowTouchEnroll === true`


dpgraham:
Published to 1.23.1

PR #108, opened 2017-06-06
Add method for dismissing possible Increase Database alert

imurchie:
AppleScript-based method to hit the "Increase Database" alert that comes up in Safari when a page requests logs of WebSQL/LocalStorage space. This button is not automatable through UI Automation or XCTest.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -16,6 +16,7 @@ import events from 'events';
 import Calendar from './calendar';
 const { EventEmitter } = events;
 
+
 const OPEN_TIMEOUT = 3000;
 const STARTUP_TIMEOUT = 60 * 1000;
 const EXTRA_STARTUP_TIME = 2000;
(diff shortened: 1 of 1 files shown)


lib/simulator-xcode-6.js
```diff
@@ -610,6 +611,14 @@ class SimulatorXcode6 extends EventEmitter {
...
+  async dismissDatabaseAlert (increase = true) {
+    let button = increase ? 'Increase' : 'Cancel';
+    log.debug(`Attempting to dismiss database alert with '${button}' button`);
+    await exec('osascript', ['-e', `
+      activate application "Simulator"
      tell application "System Events"
      	tell process "Simulator"
      		click button "${button}" of window 1
      	end tell
      end tell
```
jlipps: this is kind of weird looking, are the `^M` necessary instead of `\\\n` or something similar?

imurchie: Hmm. Yes, that is weird. Backticks should just take the newlines in stride.

PR #110, opened 2017-07-16
Make the code more compatible to multiple Simulators management

mykola-mokhnach:
We need to get rid of all the stuff, which does not work with simctl and the device udid directly (like grepping process names, killing all services by their names, and not by udid, etc), otherwise it will not be possible to isolate simulator interaction in scope of a single session and not to touch all the others parallel sessions running at the same time. We need also to think about adding headless simulator testing support.

### DIFF
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -8,9 +8,9 @@ import { getAppContainer, openUrl as simctlOpenUrl, terminate } from 'node-simct
 
 // these sims are sloooooooow
 const STARTUP_TIMEOUT = 120 * 1000;
-const SPRINGBOARD_PROCESS_NAME = 'com.apple.springboard';
-const MOBILE_SAFARI_PROCESS_NAME = 'com.apple.mobilesafari';
-const PROCESS_LAUNCH_OK_PATTERN = (name) => new RegExp(`${name.replace('.', '\\.')}:\\s+\\d+`);
+const SPRINGBOARD_BUNDLE_ID = 'com.apple.springboard';
+const MOBILE_SAFARI_BUNDLE_ID = 'com.apple.mobilesafari';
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-8.js
```diff
@@ -33,94 +37,89 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
-          // Continue iteration in case of error
+        } catch (err) {
+          lastError = err.stderr || err.message;
         }
         return false;
       }, {waitMs: startupTimeout, intervalMs: 500});
     } catch (err) {
-      log.errorAndThrow(`Simulator is not booted after ${process.hrtime(startupTimestamp)[0]} seconds`);
+      log.errorAndThrow(`Simulator is not booted after ${process.hrtime(startupTimestamp)[0]} seconds because of ` +
+                        `"${lastError ? '"' + lastError + '"' : 'an unknown error'}"`);
```
imurchie: The formatting here seems strange. Won't it print either `... because of ""whatever the error""` or `... because of "an unknown error"`?

Perhaps we need
```js
`${lastError ? `'${lastError}'` : 'an unknown error'}`);
```

mykola-mokhnach: @imurchie you can, actually, [see](https://travis-ci.org/appium/appium-ios-simulator/jobs/254210962) how the output looks like.

imurchie: So the answer is yes, the formatting is weird?

https://travis-ci.org/appium/appium-ios-simulator/jobs/254210962#L1927-L1929

mykola-mokhnach: thanks, changed

imurchie:
Published in `appium-ios-simulator@1.25.0`.

PR #111, opened 2017-07-17
Use more accurate isRunning detection

mykola-mokhnach:
The current verification may return true when the simulator is still in Booting state and is not ready to execute commands


mykola-mokhnach:
I double checked that and it looks like mine implementation is not very different from the current one :(

PR #112, opened 2017-07-17
Add headless capability support to Xcode 9 Simulator

mykola-mokhnach:
Some adjustments are still needed to be done on xcuitest driver side, but this patch is not going to break anything anyway. It would be nice to have Xcode9-beta support on Travis, so we can create a functional test for the patch.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -71,7 +71,6 @@ class SimulatorXcode6 extends EventEmitter {
     if (newValue) {
       log.warn('The headless option has no effect on Simulator below Xcode9');
     }
-    this._isHeadless = !!newValue;
   }
 
   get isHeadless () {
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -46,6 +46,7 @@ class SimulatorXcode6 extends EventEmitter {
 
     this.scaleFactor = null;
     this.connectHardwareKeyboard = null;
+    this._isHeadless = false;
 
     this.appDataBundlePaths = {};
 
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-6.js
```diff
@@ -66,6 +67,17 @@ class SimulatorXcode6 extends EventEmitter {
     this.calendar = new Calendar(this.getDir());
   }
 
+  set isHeadless (newValue) {
+    if (newValue) {
+      log.warn('The headless option has no effect on Simulator below Xcode9');
+    }
+    this._isHeadless = !!newValue;
```
imurchie: Outside of Xcode 9 `isHeadless` should always be false, no matter what is sent in to this, no?

mykola-mokhnach: yep, changed

lib/simulator-xcode-9.js
```diff
@@ -1,12 +1,58 @@
...
 
+const SIMULATOR_UI_APP_PATTERN = 'Simulator\\.app';
+const SIMULATOR_SHUTDOWN_TIMEOUT = 15 * 1000;
+
 class SimulatorXcode9 extends SimulatorXcode8 {
   constructor (udid, xcodeVersion) {
     super(udid, xcodeVersion);
   }
 
+  set isHeadless (newValue) {
```
jlipps: i know this PR is merged, but i have a question. do we expect `isHeadless` to change during the lifetime of an instance? and if not, why not just have it be another option passed into the constructor? we can still warn if necessary.

mykola-mokhnach: I didn't want to change the constructor, otherwise it might require additional changes in the client drivers. If we decide to refactor the stuff then I'd like to change scaleFactor implementation and some other additions, which only work before Simulator is booted

jlipps: ok, fine with me as is. but should consider doing some of that refactor work all at once then as you pointed out.

PR #114, opened 2017-07-20
Test with Xcode 9/iOS 11 in Travis

imurchie:
And a few little changes to make it work consistently and reliably.

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -3,6 +3,7 @@ import { exec } from 'teen_process';
 import log from './logger';
 import { shutdown as simctlShutdown, bootDevice } from 'node-simctl';
 import { waitForCondition } from 'asyncbox';
+import { restoreTouchEnrollShortcuts, backupTouchEnrollShortcuts } from './touch-enroll.js';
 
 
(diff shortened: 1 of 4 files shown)


lib/simulator-xcode-9.js
```diff
@@ -51,8 +52,14 @@ class SimulatorXcode9 extends SimulatorXcode8 {
   }
 
   async shutdown () {
+    await restoreTouchEnrollShortcuts();
```
mykola-mokhnach: here one must verify whether UI client is running first, because it might be we are running in headless mode

imurchie: Why does it matter whether it is headless or not for this operation?

mykola-mokhnach: OK, now I see that restoreTouchEnrollShortcuts has nothing to do with the UI. However it might still be an issue if we are running parallel tests

mykola-mokhnach: we need to make sure that no actions we do in the OS can affect other tests running in parallel

imurchie: Yes. As of now we have no tests, and no sense that running in parallel will work anyway. This PR is just to get the machinery in place to be able to test such things, while maintaining the current functionality.

PR #115, opened 2017-07-20
Move the documentation to docstrings and format it according to JSDoc standard

mykola-mokhnach:
The formatting rules can be found at http://usejsdoc.org/


mykola-mokhnach:
@jlipps which backend shall we select for documents generation: JSDoc3 or ESDoc or Sphinx-js? I'd be happy if you select one and help me to properly add it to the package manifest.

jlipps:
Hard to know which of these to use. Each of them has a pretty full feature set. JSDoc has what seems to be more template options. Sphinx-js has a crazy amount of features but it entails rewriting everything in rst. I think it'd be good to use whichever solution we are eventually also going to use for our tutorials and docs. So I think I'd vote for JSDoc3 or ESDoc. Mixing of manual pages and API pages seems to be a bit more natural in ESDoc but that could just be an initial impression. Probably @imurchie should weigh in here too.

imurchie:
ESDoc seems to be good.

imurchie:
Merging this now. When there is time we can figure out how to build/serve the docs.

PR #116, opened 2017-07-21
Fix Simulator startup if UI client is running under Xcode9

mykola-mokhnach:
This PR solves Simulator startup issue under Xcode9 in case if the UI client is already running (which is not the case for Xcode8 and earlier because shutdown call kills UI client there and we cannot do it under Xcode9, because it will kill all other booted Simulator instances).

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -296,7 +296,11 @@ class SimulatorXcode6 extends EventEmitter {
       await setTouchEnrollKey();
     }
     if (await this.isUIClientRunning()) {
-      await simctl.bootDevice(this.udid);
+      try {
+        await bootDevice(this.udid);
+      } catch (err) {
+        log.warn(`'xcrun simctl boot ${this.udid}' command has returned non-zero code. The problem was: ${err.stderr}`);
+      }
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-6.js
```diff
@@ -284,43 +284,43 @@ class SimulatorXcode6 extends EventEmitter {
...
-      args.push('-ConnectHardwareKeyboard', '0');
-    }
-
+    const startTime = process.hrtime();
     // Set the 'Touch ID Enroll' key bindings before the Simulator starts
     if (opts.allowTouchEnroll) {
       await setTouchEnrollKey();
     }
+    if (await this.isUIClientRunning()) {
+      await simctl.bootDevice(this.udid);
```
imurchie: Is this only in Xcode 9? Should this method be overridden there?

mykola-mokhnach: This is ok for other simulators as well, since this.shutdown() kills the UI client as well prior  toXcode9, which means this condition branch is never going to be executed there.

imurchie:
Published in `appium-ios-simulator@2.0.1`.

PR #117, opened 2017-07-21
Tune enroll backups logic not to break parallel tests

mykola-mokhnach:
Tune touch enroll tests to avoid the situation when parallel simulators unexpectedly change states of shared user defaults. This will, probably, will require some more work to prevent UI Apple scripts to be executed at the same time.

### DIFF
--- a/lib/touch-enroll.js
+++ b/lib/touch-enroll.js
@@ -92,7 +92,7 @@ async function backupTouchEnrollShortcuts () {
 async function restoreTouchEnrollShortcuts () {
   if (touchEnrollBackupsStack.length) {
     const preservedKeys = touchEnrollBackupsStack.pop();
-    if (!touchEnrollBackupsStack.length) {
+    if (preservedKeys) {
       await setTouchEnrollKeys(preservedKeys);
     }
   }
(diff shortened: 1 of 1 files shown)


lib/touch-enroll.js
```diff
@@ -65,29 +65,41 @@ async function getUserDefault (domain, key) {
...
+ * Only one backup can be preserved. Multiple calls to this method
+ * will have no effect if the backup has already been restored or hasn't been made.
+ */
 async function restoreTouchEnrollShortcuts () {
-  if (touchEnrollBackups) {
-    await setTouchEnrollKeys(touchEnrollBackups);
-    touchEnrollBackups = undefined;
+  if (touchEnrollBackupsStack.length) {
+    const preservedKeys = touchEnrollBackupsStack.pop();
+    if (!touchEnrollBackupsStack.length) {
```
imurchie: I don't understand this. Only a single element will ever be put in the stack, so once we pop either `preservedKeys` has something in it, or is `undefined` (if you `pop` from the empty stack). Shouldn't that be checked, instead of the `length`, which can only be `0`?

mykola-mokhnach: yep, that would be more readable

PR #119, opened 2017-07-28
Ensure the correct simulator is booted while UI client is running

mykola-mokhnach:
It looks like newer releases of Simulator UI client may ignore some arguments to start the particular client. That is why we try boot the correct server in any case independently of the UI state.


imurchie:
Published in `appium-ios-simulator@2.1.1`.

PR #121, opened 2017-08-02
Improve UI client state detection

mykola-mokhnach:
It looks like the current approach is not very stable in parallel tests (it always tries to start the UI client even if it is already running).

### DIFF
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -11,6 +11,7 @@ const STARTUP_TIMEOUT = 120 * 1000;
 const SAFARI_STARTUP_TIMEOUT = 25 * 1000;
 const SPRINGBOARD_BUNDLE_ID = 'com.apple.springboard';
 const MOBILE_SAFARI_BUNDLE_ID = 'com.apple.mobilesafari';
+const UI_CLIENT_BUNDLE_ID = 'com.apple.iphonesimulator';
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
Looking for a best method to check if UI client is running

mykola-mokhnach:
Found better way to detect UI client running state

lib/simulator-xcode-8.js
```diff
@@ -29,6 +30,49 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
+
+  /**
+   * Check the state of Simulator UI client.
+   * @Override
+   *
+   * @return {boolean} True of if UI client is running or false otherwise.
+   */
+  async isUIClientRunning () {
+    const args = ['-e', `tell application "System Events" to count processes whose bundle identifier is "${this.uiClientBundleId}"`];
+    const {stdout} = await exec('osascript', args);
```
imurchie: Does this need any special permissions for the terminal (or whatever is running the Appium process)?

imurchie: Seems like my answer is "no", since Travis is fine.

mykola-mokhnach: I suppose not, because we don't interact with application UI.

yaphatak: why we cant find process `ps -ax|grep -i 'udid'|grep -i 'Simulator'` like this and kill it

mykola-mokhnach: @yaphatak we could, but I wanted more precise method. Application bundle id is unique and the process name can have duplicates.

PR #122, opened 2017-08-03
Add a separate test for headless mode

mykola-mokhnach:
check headless mode functionality on Xcode9+

### DIFF
--- a/test/simulator-e2e-specs.js
+++ b/test/simulator-e2e-specs.js
@@ -268,6 +268,36 @@ function runTests (deviceType) {
       (await sim.hasCalendarAccess(BUNDLE_ID)).should.be.false;
     });
 
+    it('should properly start simulator in headless mode on Xcode9+', async function () {
+      const sim = await getSimulator(udid);
+      if (!(parseInt(sim.constructor.name.replace(/\D/g, ''), 10) >= 9)) {
+        // This feature is only supported on XCode9+
+        this.skip();
+      }
+
+      const verifyStates = async (shouldServerRun, shouldClientRun) => {
+        const isServerRunning = await sim.isRunning();
+        isServerRunning.should.eql(shouldServerRun);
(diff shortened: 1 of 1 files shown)


test/simulator-e2e-specs.js
```diff
@@ -260,6 +260,36 @@ function runTests (deviceType) {
       (await sim.hasCalendarAccess(BUNDLE_ID)).should.be.false;
     });
 
+    it('should properly start simulator in headless mode on Xcode9+', async function () {
+      const sim = await getSimulator(udid);
+      if (!(parseInt(sim.constructor.name.replace(/\D/g, ''), 10) >= 9)) {
```
mykola-mokhnach: Is there a better way to detect the current xcode version here?

imurchie: Yes. We already have `appium-xcode` as a dependency. So you can just use something like:
```js
import xcode from 'appium-xcode';

// ...

before(async function () {
  let xcodeVersion = await xcode.getVersion(true);
  if (xcodeVersion.major < 9) return this.skip();
});
```

test/simulator-e2e-specs.js
```diff
@@ -260,6 +261,36 @@ function runTests (deviceType) {
       (await sim.hasCalendarAccess(BUNDLE_ID)).should.be.false;
     });
 
+    it('should properly start simulator in headless mode on Xcode9+', async function () {
+      const sim = await getSimulator(udid);
```
imurchie: These should be revered. There is no reason to have the sim object if we are going to skip.

mykola-mokhnach:
@imurchie done. Also moved verification method to the top, because I'm going to reuse it for parallel session tests (https://github.com/appium/appium-ios-simulator/pull/123)

imurchie:
Awesome! Yes. I noted that on the other PR :)

PR #123, opened 2017-08-03
Add functional test for parallel simulators startup

mykola-mokhnach:
That's unexpected, but it works

### DIFF
--- a/test/simulator-e2e-specs.js
+++ b/test/simulator-e2e-specs.js
@@ -381,46 +381,32 @@ function runTests (deviceType) {
       }
 
       await killAllSimulators();
-      for (const deviceIdx of _.range(1, DEVICES_COUNT + 1)) {
-        const udid = await simctl.createDevice(`ios-simulator_${deviceIdx}_testing`,
-                                               deviceType.device,
-                                               deviceType.version);
-        const sim = await getSimulator(udid);
-        simulatorsMapping.set(udid, sim);
-      }
+      const udids = await B.map(_.range(1, DEVICES_COUNT + 1),
+                                (deviceIdx) => simctl.createDevice(`ios-simulator_${deviceIdx}_testing`,
+                                                                   deviceType.device,
+                                                                   deviceType.version));
+      const simulators = await B.map(udids, (udid) => getSimulator(udid));
+      _.zip(udids, simulators).map(([udid, sim]) => simulatorsMapping.set(udid, sim));
     });
     afterEach(async function () {
-      await killAllSimulators();
-      const existingDevices = (await simctl.getDevices())[deviceType.version];
-      if (existingDevices) {
-        for (const existingDevice of existingDevices) {
-          if (simulatorsMapping.has(existingDevice.udid)) {
-            try {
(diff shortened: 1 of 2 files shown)


test/simulator-e2e-specs.js
```diff
@@ -360,6 +361,74 @@ function runTests (deviceType) {
       await getTouchEnrollKeys().should.eventually.deep.equal(originalValues);
     });
   });
+
+  describe(`multiple instances of ${deviceType.version} simulator on Xcode9+`, function () {
+    this.timeout(LONG_TIMEOUT);
+    const simulatorsMapping = new Map();
```
imurchie: Why not just a normal old object?

test/simulator-e2e-specs.js
```diff
@@ -360,6 +361,74 @@ function runTests (deviceType) {
...
+    const DEVICES_COUNT = 2;
+    const verifyStates = async (sim, shouldServerRun, shouldClientRun) => {
```
imurchie: This will eventually have to be pulled out to a more accessible place, since the headless test also uses it.

test/simulator-e2e-specs.js
```diff
@@ -360,6 +361,74 @@ function runTests (deviceType) {
...
+        simulatorsMapping.set(udid, sim);
+      }
+    });
+    after(async function () {
+      const xcodeVersion = await xcode.getVersion(true);
+      if (xcodeVersion.major < 9) {
+        return this.skip();
+      }
+
+      const existingDevices = (await simctl.getDevices())[deviceType.version];
```
imurchie: It's unclear to me what is gained by this. Can't you just iterate through the keys in `simulatorsMapping` and delete everything?

test/simulator-e2e-specs.js
```diff
@@ -360,6 +361,74 @@ function runTests (deviceType) {
...
+          try {
+            await simctl.deleteDevice(existingDevice.udid);
+          } catch (ign) {}
+        }
+      }
+      simulatorsMapping.clear();
+    });
+
+    it('should start multiple simulators in "default" mode', async function () {
+      const xcodeVersion = await xcode.getVersion(true);
```
imurchie: This should not be necessary if you skip in the `before` block.

jlipps:
LGTM, @imurchie has some good questions though

test/simulator-e2e-specs.js
mykola-mokhnach: it's to make sure all the devices, we tried to create in _before_ method, exist and ready for removal. There is a possibility they don't exist if _before_ method failed.

test/simulator-e2e-specs.js
mykola-mokhnach: because I use the _clear_ method, which plain object doesn't have. Also i find this data type more specialised for such type of tasks. Objects are too generic.

test/simulator-e2e-specs.js
mykola-mokhnach: removed

PR #125, opened 2017-08-08
Avoid stopping simulator services before erase

mykola-mokhnach:
It looks like stopping of Simulator services prevents simulator to be properly erased on Xcode9+.


imurchie:
Published in `appium-ios-simulator@2.2.1`.

PR #126, opened 2017-08-10
Simplify touch ID enrollment  procedure

mykola-mokhnach:
The current implementation was not working nicely with parallel sessions. Now this should be fixed


test/functional/touch-enroll-e2e-specs.js
```diff
@@ -1,164 +0,0 @@
-// transpile:mocha
```
jlipps: why do we want to get rid of all these user defaults tests?

PR #127, opened 2017-08-11
Resume e2e tests on Travis

imurchie:


package.json
```diff
@@ -38,22 +38,25 @@
...
-    "lint",
     "test"
   ],
   "devDependencies": {
-    "appium-gulp-plugins": "^2.0.0",
+    "appium-gulp-plugins": "^2.1.4",
     "babel-eslint": "^7.1.1",
-    "chai": "^4.1.0",
-    "chai-as-promised": "^7.1.1",
+    "chai": "^3.2.0",
```
mykola-mokhnach: do we really want to use older versions?

mykola-mokhnach:
[Here](https://travis-ci.org/appium/appium-ios-simulator/jobs/263654661) I see two failed tests, but the overall result is Passed. Should it be like this?

imurchie:
I saw that too. I have to figure out why the reporting is wrong.

PR #128, opened 2017-09-01
Fix system apps detection

mykola-mokhnach:
The PR fixes system applications detection on Simulator. Please do not merge it until https://github.com/appium/node-simctl/commit/a5c4bfb33fcaf5139cbfeccc62c3dc49d005f06c is published

### DIFF
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -3,7 +3,7 @@ import SimulatorXcode7 from './simulator-xcode-7';
 import log from './logger';
 import { waitForCondition } from 'asyncbox';
 import { exec } from 'teen_process';
-import { getAppContainer, openUrl as simctlOpenUrl, terminate } from 'node-simctl';
+import { getAppContainer, openUrl as simctlOpenUrl, terminate, appInfo } from 'node-simctl';
 
 
 // these sims are sloooooooow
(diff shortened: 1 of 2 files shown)


imurchie:
`node-simctl@3.11.0` has the fix.

lib/simulator-xcode-8.js
```diff
@@ -94,15 +94,22 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
-      let appContainer = await getAppContainer(this.udid, bundleId, false);
+      const appContainer = await getAppContainer(this.udid, bundleId, false);
       return appContainer.endsWith('.app');
     } catch (err) {
-      return false;
+      // get_app_container subcommand fails for system applications,
+      // so we try the hidden appinfo subcommand, which prints correct info for
+      // system/hidden apps
+      try {
+        const appInfo = await appInfo(this.udid, bundleId);
```
imurchie: Does this command _not_ work for regular apps? Why not use it for everything?

mykola-mokhnach: it does work, but the fact that this command is hidden makes me a bit suspicious about it. The stuff might stop working completely Apple decides to change the behaviour there in some future releases. Although the current approach will still work for non-system apps, which is the most common clients case.

imurchie:
Published in `appium-ios-simulator@2.2.5`.

PR #131, opened 2017-10-01
Add possibility to set simulator window center

vrunoa:
Prepare appium-ios-simulator to receive simWindowCenter capability to set the simulator position on the screen. This is only available for ios11, @mykola-mokhnach wanna share some thoughts on how to validate this on the _run_ method ? Or should we validate it during capability parsing on appium-ios-driver ?


mykola-mokhnach:
I'd rather pass _simulatorWindowCenter_ string value directly from xcuitest driver, so you could simply perform an additional value format verification here using a regexp.

lib/simulator-xcode-9.js
```diff
@@ -64,6 +66,11 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     if (opts.scaleFactor) {
       opts.devicePreferences.SimulatorWindowLastScale = parseFloat(opts.scaleFactor);
     }
+    if (opts.simWindowCenter && opts.simWindowCenter.x && opts.simWindowCenter.y) {
+      opts.devicePreferences.SimulatorWindowCenter = `{${opts.simWindowCenter.x},${opts.simWindowCenter.y}}`;
+    } else {
+      log.warn(`Invalid simWindowCenter value: ${opts.simWindowCenter}. Valid format {x:x, y:y}`);
```
mykola-mokhnach: this warning will be always shown even if _simWindowCenter_ is not set. I don't think this is the expected behaviour.

vrunoa:
@mykola-mokhnach I feel like it's easier to understand and to maintain if params are treated like an object with coordinates.

mykola-mokhnach:
You'll have to perform double transformation then, since the plist anyway requires a string value. The initial idea was that a string value is passed from xcuitest driver and then you just verify it has an appropriate format before writing it to the file. This could be done in verifyDevicePreferences method, which has been added in https://github.com/appium/appium-ios-simulator/pull/132

lib/simulator-xcode-9.js
```diff
@@ -64,6 +66,14 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     if (opts.scaleFactor) {
       opts.devicePreferences.SimulatorWindowLastScale = parseFloat(opts.scaleFactor);
     }
+    if (opts.simulatorWindowCenter) {
+      if (!isNaN(opts.simulatorWindowCenter.x) && !isNaN(opts.simulatorWindowCenter.y)) {
+        let x = parseFloat(opts.simulatorWindowCenter.x), y = parseFloat(opts.simulatorWindowCenter.y);
+        opts.devicePreferences.SimulatorWindowCenter = `{${x},${y}}`;
+      } else {
+        log.warn(`Invalid simWindowCenter value: ${opts.simWindowCenter}. Valid format {x:x, y:y}`);
```
imurchie: Is it `opts.simWindowCenter` or `opts.simulatorWindowCenter`.

lib/simulator-xcode-9.js
```diff
@@ -64,6 +66,14 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     if (opts.scaleFactor) {
       opts.devicePreferences.SimulatorWindowLastScale = parseFloat(opts.scaleFactor);
     }
+    if (opts.simulatorWindowCenter) {
+      if (!isNaN(opts.simulatorWindowCenter.x) && !isNaN(opts.simulatorWindowCenter.y)) {
+        let x = parseFloat(opts.simulatorWindowCenter.x), y = parseFloat(opts.simulatorWindowCenter.y);
+        opts.devicePreferences.SimulatorWindowCenter = `{${x},${y}}`;
+      } else {
+        log.warn(`Invalid simWindowCenter value: ${opts.simulatorWindowCenter}. Valid format {x:x, y:y}`);
```
imurchie: I think the name printed out ought to be the correct name, too, no?

mykola-mokhnach:
I don't really like the current implementation, since it is easier and more readable to pass 

```
setCapability("simulatorWindowCenter", "{100.0, -100.0}");
```

in comparison to 

```
setCapability("simulatorWindowCenter", "{\"x\":100.0, \"y\":-100.0}");
```

vrunoa:
@mykola-mokhnach why would you set it as a string, why cant be an object ?

```javascript
setCapability("simulatorWindowCenter", {x:100.0, y:100.0});
```

mykola-mokhnach:
Because capability type can be only a primitive one: string, boolean, number

vrunoa:
@mykola-mokhnach changed. I'll add some unit tests.

lib/simulator-xcode-9.js
```diff
@@ -6,6 +6,7 @@ import AsyncLock from 'async-lock';
 import log from './logger';
 import { shutdown as simctlShutdown, bootDevice, eraseDevice } from 'node-simctl';
 import { waitForCondition } from 'asyncbox';
+import { validCoordinates } from './utils';
```
mykola-mokhnach: it is not necessary to create a separate utility method. Just merge with the recent master and put one more verification into _verifyDevicePreferences_ method

lib/simulator-xcode-9.js
```diff
@@ -54,6 +55,8 @@ class SimulatorXcode9 extends SimulatorXcode8 {
    *   client invisible). `false` by default.
    *   - {DevicePreferences} devicePreferences: preferences of the newly created Simulator
    *   device
+   *   - {object} simulatorWindowCenter: object with x,y coordinates to place the Simulator window
```
mykola-mokhnach: this docstring should be updated now

mykola-mokhnach: you can, actually, set this property directly to opts.devicePreferences,
_scaleFactor_ is set separately because of legacy reasons

lib/simulator-xcode-9.js
```diff
@@ -64,6 +67,13 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     if (opts.scaleFactor) {
       opts.devicePreferences.SimulatorWindowLastScale = parseFloat(opts.scaleFactor);
     }
+    if (opts.simulatorWindowCenter) {
```
mykola-mokhnach: I still would prefer this option to be written directly into _devicePreferences_

lib/simulator-xcode-9.js
```diff
@@ -64,6 +67,13 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+      if (validCoordinates(opts.simulatorWindowCenter)) {
+        opts.devicePreferences.SimulatorWindowCenter = opts.simulatorWindowCenter;
+      } else {
+        log.warn(`Invalid simulatorWindowCenter value: ${opts.simulatorWindowCenter}. Valid string format "{x,y}"`);
```
mykola-mokhnach: ... and the verification to be moved into _verifyDevicePreferences_ method

jlipps:
@mykola-mokhnach i didn't realize capability values can't be objects. where are you seeing that? passing in a string that looks like a JSON object but isn't is bound to be confusing, IMO.

mykola-mokhnach:
@jlipps There is _isObject_ validator, however I didn't see it's used anywhere as capability type.
Also, it's unclear how to pass it from different clients. I mean it's ok to pass a JSON from wd.io, but how about java/python/ruby/php? That is why I'd prefer having a simple string and regexp verification. Also, the type of the original plist value is also String

jlipps:
@mykola-mokhnach ok good point. At the end of the day desired capabilities get stringified as a JSON string, so it should work. But you are right that different clients may not make it easy to set a value of a desired capability as an object. So I'm fine with the current approach.

lib/simulator-xcode-9.js
vrunoa: @mykola-mokhnach where's the _verifyDevicePreferences_ method ?

mykola-mokhnach: just rebase with the master branch

mykola-mokhnach:
I close this Pr, since the corresponding functionality has been implemented as part of https://github.com/appium/appium-ios-simulator/pull/134

@vrunoa Do you have time to add _simulatorWindowCenter_ capability to xcuitest driver?

vrunoa:
hey @mykola-mokhnach sorry, I've been pretty busy the last few weeks. I could take a look next week, i'm going to be quite free.

mykola-mokhnach:
np, I could do it if you don't mind

vrunoa:
go ahead. I'll try to close another issue i've been working on next week. thanks

PR #134, opened 2017-10-19
Apply validation rules for device preferences

mykola-mokhnach:
After this PR is merged it will only be necessary to add and document the corresponding capability for xcui driver
FYI @vrunoa

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -152,6 +152,7 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     }
 
     if (!_.isUndefined(prefs.SimulatorWindowCenter)) {
+      // https://regex101.com/r/2ZXOij/2
       const verificationPattern = /\{\-?\d+(\.\d+)?,\-?\d+(\.\d+)?\}/;
       if (!_.isString(prefs.SimulatorWindowCenter) || !verificationPattern.test(prefs.SimulatorWindowCenter)) {
         log.errorAndThrow(`SimulatorWindowCenter is expected to match "{floatXPosition,floatYPosition}" format (without spaces). ` +
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -150,6 +150,29 @@ class SimulatorXcode9 extends SimulatorXcode8 {
           `'${prefs.SimulatorWindowLastScale}' is assigned instead.`);
       }
     }
+
+    if (!_.isUndefined(prefs.SimulatorWindowCenter)) {
+      const verificationPattern = /\{\-?\d+(\.\d+)?,\-?\d+(\.\d+)?\}/;
+      if (!_.isString(prefs.SimulatorWindowCenter) || !verificationPattern.test(prefs.SimulatorWindowCenter)) {
+        log.errorAndThrow(`SimulatorWindowCenter is expected to match "{floatXPosition,floatYPosition}" format (without spaces). ` +
+          `'${prefs.SimulatorWindowCenter}' is assigned instead.`);
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-9.js
```diff
@@ -150,6 +150,29 @@ class SimulatorXcode9 extends SimulatorXcode8 {
           `'${prefs.SimulatorWindowLastScale}' is assigned instead.`);
       }
     }
+
+    if (!_.isUndefined(prefs.SimulatorWindowCenter)) {
+      const verificationPattern = /\{\-?\d+(\.\d+)?,\-?\d+(\.\d+)?\}/;
```
imurchie: Can you document this regex in some way? Perhaps in https://regex101.com/ or something?

lib/simulator-xcode-9.js
```diff
@@ -150,6 +150,29 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+      if (!_.isString(prefs.SimulatorWindowCenter) || !verificationPattern.test(prefs.SimulatorWindowCenter)) {
+        log.errorAndThrow(`SimulatorWindowCenter is expected to match "{floatXPosition,floatYPosition}" format (without spaces). ` +
```
imurchie: Why not accommodate spaces?

mykola-mokhnach: because there are no spaces in the original plist, so I'm not sure if it works properly if I add some

lib/simulator-xcode-9.js
mykola-mokhnach: would it be ok https://regex101.com/r/2ZXOij/1 ?

imurchie: Awesome!

This points out, though, that the capture groups are unnecessary, since nothing is being done with what is captured (`/\{\-?\d+\.\d+?,\-?\d+\.\d+?\}/` would suffice). I don't think it matters here because the efficiency is not really at issue for such a short search. But in general capturing text is much less performant in a regex.

lib/simulator-xcode-9.js
imurchie: Ok. It might be good to allow them as input (even if the spaces need to be stripped out on the server side) since some language's serialization library might include spaces at some point. Perhaps that is better left for if/when such an event occurs. From the error message it would be obvious what the problem is.

lib/simulator-xcode-9.js
mykola-mokhnach: I just wanted to make sure the expression like this is not possible
`{1., 2.}`
Would it be possible without groups?

lib/simulator-xcode-9.js
mykola-mokhnach: I expect the value to be passed as a simple string, so nothing is going to be changed by the language

lib/simulator-xcode-9.js
imurchie: Ah. Yes, you need them for that. Can you put the various variations in the linked regex test (you'll need to add the `m` flag so it is multiline there?
```
# should match
{100.0,100.0}
{100,100}
{100.0,100}

# should not match
{100.0, 100.0}
{100.0,100.}
```

lib/simulator-xcode-9.js
imurchie: But if, in my test, I compute the value, put it in an object, and then serialize that, if the serialization adds a space it would not work.

But this is conjecture. I don't know of any particular JSON serializer that would do it, and generally they like to compress as much as possible.

lib/simulator-xcode-9.js
mykola-mokhnach: I thought a unit test would be enough for this purpose. But yeah, it's easy to add

imurchie:
Published in `appium-ios-simulator@2.4.0`

PR #135, opened 2017-10-22
Add a possibility to clear Simulator caches

mykola-mokhnach:
The next step will be to include the cleanup of `com.apple.mobile.installd.staging` cache folder into xcuitest driver lgorithm. See https://github.com/appium/appium/issues/9410 for more details.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -776,7 +776,7 @@ class SimulatorXcode6 extends EventEmitter {
    */
   async clearCaches (...folderNames) {
     const cachesRoot = path.resolve(this.getDir(), 'Library', 'Caches');
-    if (!await fs.hasAccess(cachesRoot)) {
+    if (!(await fs.hasAccess(cachesRoot))) {
       log.debug(`Caches root at '${cachesRoot}' does not exist or is not accessible. Nothing to do there`);
       return 0;
     }
@@ -794,7 +794,11 @@ class SimulatorXcode6 extends EventEmitter {
 
     log.debug(`Matched ${itemsToRemove.length} Simulator cache ` +
       `item${itemsToRemove.length === 1 ? '' : 's'} for cleanup: ${itemsToRemove}`);
-    await B.all(itemsToRemove, (x) => fs.rimraf(x));
+    try {
+      await B.all(itemsToRemove, (x) => fs.rimraf(x));
+    } catch (e) {
+      log.warn(`Got an exception while cleaning Simulator caches: ${e.message}`);
+    }
     return itemsToRemove.length;
   }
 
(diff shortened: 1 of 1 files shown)


lib/simulator-xcode-6.js
```diff
@@ -450,7 +450,7 @@ class SimulatorXcode6 extends EventEmitter {
    * @param {boolean} scrub - If `scrub` is false, we want to clean by deleting the app and all
    *   files associated with it. If `scrub` is true, we just want to delete the preferences and
    *   changed files.
-   * @return {array} Array of deletion promises.
+   * @return {array<Promise>} Array of deletion promises.
```
imurchie: This method ends in `await B.all(deletePromises);`. So it does not return anything at all.

lib/simulator-xcode-6.js
```diff
@@ -436,10 +436,10 @@ class SimulatorXcode6 extends EventEmitter {
    *
    * @param {string} appFile - Application name minus ".app".
    * @param {string} appBundleId - Bundle identifier of the application.
-   * @return {array} Array of deletion promises.
+   * @return {array<Promise>} Array of deletion promises.
    */
   async scrubCustomApp (appFile, appBundleId) {
-    return await this.cleanCustomApp (appFile, appBundleId, true);
+    return await this.cleanCustomApp(appFile, appBundleId, true);
```
imurchie: See comment below.

lib/simulator-xcode-6.js
```diff
@@ -768,6 +768,36 @@ class SimulatorXcode6 extends EventEmitter {
...
+    if (!await fs.hasAccess(cachesRoot)) {
+      log.debug(`Caches root at '${cachesRoot}' does not exist or is not accessible. Nothing to do there`);
+      return this;
+    }
+
+    let itemsToRemove = folderNames.length ?
+      folderNames.map((x) => path.resolve(cachesRoot, x)) :
+      (await fs.readdir(cachesRoot)).map((x) => path.resolve(cachesRoot, x));
+    itemsToRemove = await B.filter(itemsToRemove, (x) => fs.hasAccess(x));
+    itemsToRemove = await B.filter(itemsToRemove, async (x) => (await fs.stat(x)).isDirectory());
```
imurchie: These could all be chained if you wanted (since everything returns a Promise.

imurchie: Which is more likely to filter more things, in the normal case? The order might need adjusting if there are more non-directories than directories that cannot be accessed.

lib/simulator-xcode-6.js
mykola-mokhnach: fixed

lib/simulator-xcode-6.js
mykola-mokhnach: fixed

lib/simulator-xcode-6.js
mykola-mokhnach: moved isDirectory check to the top

imurchie:
Published in `appium-ios-simulator@2.5.0`.

PR #136, opened 2017-10-24
Add a possibility to perform Shake gesture and set custom geolocation coordinates via AppleScript

mykola-mokhnach:
Also did some refactoring to improve the algorithm for multi-window configuration.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -859,6 +859,22 @@ class SimulatorXcode6 extends EventEmitter {
   }
 
   /**
+   * Generate Apple script, which activates Simulator.
+   *
+   * @returns {string} the generated Apple Script snipped.
+   */
+  async generateWindowActivationScript () {
+    return `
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-6.js
```diff
@@ -859,6 +859,22 @@ class SimulatorXcode6 extends EventEmitter {
   }
 
   /**
+   * Generate Apple script, which activates Simulator.
+   *
+   * @returns {string} the generated Apple Script snipped.
+   */
+  async generateWindowActivationScript () {
```
imurchie: Why make this `async`?

imurchie: Oh, I guess because a subclass needs it to be?

mykola-mokhnach: yes, exactly

imurchie:
Published in `appium-ios-simulator@2.6.0`.

PR #137, opened 2017-10-30
Clean up stat function

imurchie:
There is no reason to create a structure with all the info in it by iterating over all the devices, and then searching inside it (by iterating over the structure). 

We are already iterating over all the devices!

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -242,19 +242,16 @@ class SimulatorXcode6 extends EventEmitter {
    * }
    */
   async stat () {
-    let devices = await simctl.getDevices();
-    let normalizedDevices = [];
-
-    // add sdk attribute to all entries, add to normalizedDevices
-    for (let [sdk, deviceArr] of _.toPairs(devices)) {
-      deviceArr = deviceArr.map((device) => {
-        device.sdk = sdk;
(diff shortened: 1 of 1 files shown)


lib/simulator-xcode-6.js
```diff
@@ -242,19 +242,16 @@ class SimulatorXcode6 extends EventEmitter {
...
-    for (let [sdk, deviceArr] of _.toPairs(devices)) {
-      deviceArr = deviceArr.map((device) => {
-        device.sdk = sdk;
-        return device;
-      });
-      normalizedDevices = normalizedDevices.concat(deviceArr);
+    for (let [sdk, deviceArr] of _.toPairs(await simctl.getDevices())) {
+      for (let device of deviceArr) {
+        if (device.udid === this.udid) {
+          device.sdk = sdk;
```
mykola-mokhnach: is that ok, that we modify this object? Perhaps it would be safer to create a copy of it?

imurchie: It only exists in this `for` loop. Not sure why we would need to make _another_ copy of it. And `simctl.getDevices()` is not memoized, so it is getting a new list every time it is called, so even the object we're iterating over only exists for the outer `for` loop.

PR #138, opened 2017-12-05
Add a method to clean the apps-related keychain

Helias-Criouet:
Hello,

The application we are interested in testing has recently started storing data in the phone keychain. This results in persisting data across test sessions (e.g. a user is still logged in after uninstalling and reinstalling the app). However, we would like to remove this data in between test sessions. 

For the record, we do not want to remove every file in the keychain folder because we have a SSL certificate there that we would like to keep.
Also, we could not use the `fullReset` capability combined with the `customSSLCert` capability because we noticed the certificate was installed before the reset and was therefore deleted.
In any case, we would prefer not to destroy the simulator in between test sessions.

In the end, we found thanks to [this SO issue](https://stackoverflow.com/questions/41862997/ios-simulator-view-content-of-keychain) that removing a certain file in the simulator's keychain directory did the trick. Therefore, creating a method to remove this file and clean only the apps-related keychain seemed to be the best approach for our problem.
The idea is to then call it with `appium-xcuitest-driver` when a certain capability e.g. `keepKeychain` is set to `false`.

Any suggestion/improvement is welcome!

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -479,6 +479,13 @@ class SimulatorXcode6 extends EventEmitter {
   }
 
   /**
+   * Remove the apps-related keychain.
+   */
+  async cleanAppsKeychain () {
+    await fs.rimraf(path.resolve(this.keychainPath, 'keychain-2-debug.db'));
+  }
+
+  /**
    * Retrieve paths to dirs where application data is stored. iOS 8+ stores app data in two places,
    * and iOS 7.1 has only one directory
    *


lib/simulator-xcode-6.js
```diff
@@ -478,6 +478,13 @@ class SimulatorXcode6 extends EventEmitter {
     await B.all(deletePromises);
   }
 
+  /**
+   * Remove the apps-related keychain.
+   */
+  async cleanAppsKeychain () {
+    await fs.rimraf(path.resolve(this.keychainPath, 'keychain-2-debug.db'));
```
mykola-mokhnach: is this name always the same? perhaps it would be better to use glob expression to match it

Helias-Criouet: [This SO issue](https://stackoverflow.com/questions/41862997/ios-simulator-view-content-of-keychain) and [this other SO issue](https://stackoverflow.com/questions/1636506/how-to-see-the-keychain-for-iphone-simulator) seem to indicate it. The second one points out in the comments that this path is only valid starting with the iOS 8 SDK (that comes with XCode 6). Removing this file certainly worked for us.

Would you prefer `*.db` instead of `keychain-2-debug.db`?

mykola-mokhnach: `keychain*.db` looks better

mykola-mokhnach:
Please also sign the CLA

mykola-mokhnach:
I assume it also makes sense to update the `clean` method (actually, add an argument there), so we can preserve keychains while doing simulator reset if the `keepKeyChains` capability is set to `true`

Helias-Criouet:
Actually, after thinking a bit more about it, I came to realize that this PR is actually not truly related to the `keepKeyChain` capability.
The `keepKeyChain` capability is basically about keeping or trashing the whole `Keychains` directory (including e.g. SSL certificates info).

This PR is about only removing the file where app-related data has been stored. So I wonder if this file should not simply be included in the stuff that is deleted in the `cleanCustomApp` method. What do you think?

mykola-mokhnach:
@imurchie Do you have any ideas about how to implement keychain cleanup in the best way?

usmonster:
(bump)

mykola-mokhnach:
@Helias-Criouet Can you please check if https://github.com/appium/appium-xcuitest-driver/pull/627 works for you?

Helias-Criouet:
@mykola-mokhnach Thank you so much for the PR, the specification of the new capability looks great!

I tested multiple cases, and I only encountered one issue: I found that using a filename pattern such as `keychain*.db*` was not working for some reason. I had to prepend a '*' to my filename pattern, e.g. `*keychain*.db*` for it to work, even though the filename that I wanted to match was `keychain-2-debug.db`. 
I'm guessing [this method](https://github.com/appium/appium-ios-simulator/blob/3e244ed2fca9933dd0147cea9fc5b7906d71721a/lib/simulator-xcode-6.js#L997) is responsible.

PR #140, opened 2017-12-14
Do not raise an error if preferences plist update fails

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/9792

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -212,12 +212,19 @@ class SimulatorXcode9 extends SimulatorXcode8 {
       newPrefs.DevicePreferences = {[this.udid.toUpperCase()]: devicePrefs};
     }
     newPrefs = _.merge(newPrefs, commonPrefs);
+    let isUpdateSuccessful = true;
     await preferencesPlistGuard.acquire(SimulatorXcode9.name, async () => {
-      const currentPlistContent = await plist.parsePlistFile(plistPath);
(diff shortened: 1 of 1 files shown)


lib/simulator-xcode-9.js
```diff
@@ -212,12 +212,19 @@ class SimulatorXcode9 extends SimulatorXcode8 {
       newPrefs.DevicePreferences = {[this.udid.toUpperCase()]: devicePrefs};
     }
     newPrefs = _.merge(newPrefs, commonPrefs);
+    let isUpdateSuccessful = true;
     await preferencesPlistGuard.acquire(SimulatorXcode9.name, async () => {
```
imurchie: The return value of the passed-in function is returned by `preferencesPlistGuard.acquire` so this can be simplified to 
```js
return await preferencesPlistGuard.acquire(SimulatorXcode9.name, async () => {
  try {
    const currentPlistContent = await plist.parsePlistFile(plistPath);
    await plist.updatePlistFile(plistPath, _.merge(currentPlistContent, newPrefs), true);
    log.debug(`Updated ${this.udid} Simulator preferences at '${plistPath}' with ${JSON.stringify(newPrefs)}`);
    return true;
  } catch (e) {
    log.warn(`Cannot update ${this.udid} Simulator preferences at '${plistPath}'. ` +
             `Try to delete the file manually in order to reset it. Original error: ${e.message}`);
    return false;
  }
});
```

mykola-mokhnach: thanks, fixed

PR #141, opened 2018-01-29
fix: call 'terminate safari' only for booted simulators

kool79:
terminate safari only on booted simulators

this fix removes error 
[simctl] Error: simctl error running 'terminate': 
from log when we try to call 'terminate' on non-booted simulator.
See log example in https://github.com/appium/appium/issues/9954#issuecomment-360337526

### DIFF
--- a/lib/simulator-xcode-8.js
+++ b/lib/simulator-xcode-8.js
@@ -192,7 +192,7 @@ class SimulatorXcode8 extends SimulatorXcode7 {
    */
   async cleanSafari (keepPrefs = true) {
     try {
-      if (this.isRunning()) {
+      if (await this.isRunning()) {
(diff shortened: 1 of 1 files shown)


lib/simulator-xcode-8.js
```diff
@@ -192,7 +192,9 @@ class SimulatorXcode8 extends SimulatorXcode7 {
    */
   async cleanSafari (keepPrefs = true) {
     try {
-      await terminate(this.udid, MOBILE_SAFARI_BUNDLE_ID);
+      if (this.isRunning()) {
```
mykola-mokhnach: await is missing

kool79:
fixed

PR #143, opened 2018-02-06
Add 'hasSSLCert' to API

dpgraham:
* Exposed 'hasSSLCert' to API which checks new pemText to check if it already exists as an entry in the Trust Store
* Updated the certificate.has method to check fingerprint too
  * If the subj is there but fingerprint is different, then it returns false
  * If subj and sha1 are unchanged, then it returns true
  * This is so that we can check if a cert is already installed, and if it is, we don't have to restart simulator

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,8 @@
 // transpile:main
 
 import { getSimulator, getDeviceString } from './lib/simulator';
-import { killAllSimulators, endAllSimulatorDaemons, simExists, installSSLCert, uninstallSSLCert } from './lib/utils';
+import { killAllSimulators, endAllSimulatorDaemons, simExists, installSSLCert, uninstallSSLCert, hasSSLCert } from './lib/utils';
 import { BOOT_COMPLETED_EVENT } from './lib/simulator-xcode-6';
 
-export { getSimulator, getDeviceString, killAllSimulators, endAllSimulatorDaemons, simExists, installSSLCert, uninstallSSLCert,
+export { getSimulator, getDeviceString, killAllSimulators, endAllSimulatorDaemons, simExists, installSSLCert, uninstallSSLCert, hasSSLCert,
          BOOT_COMPLETED_EVENT };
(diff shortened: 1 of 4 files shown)


lib/certificate.js
```diff
@@ -39,7 +39,16 @@ class Certificate {
   async has (dir) {
     let subject = await this.getSubject(this.pemFilename);
     let trustStore = new TrustStore(dir);
-    return await trustStore.hasRecords(subject);
+
```
mykola-mokhnach: would it be possible to add better docstring description to this method and also other methods in this class?

index.js
```diff
@@ -1,8 +1,8 @@
 // transpile:main
 
 import { getSimulator, getDeviceString } from './lib/simulator';
-import { killAllSimulators, endAllSimulatorDaemons, simExists, installSSLCert, uninstallSSLCert } from './lib/utils';
+import { killAllSimulators, endAllSimulatorDaemons, simExists, installSSLCert, uninstallSSLCert, hasSSLCert } from './lib/utils';
```
mykola-mokhnach: these export/import lines are quite long. I'd rather split them into two rows

lib/certificate.js
```diff
@@ -39,7 +39,16 @@ class Certificate {
...
+    // Return false if record with this subject is not found
+    if (!await trustStore.hasRecords(subject)) {
+      return false;
+    }
+
+    // If record is found, check fingerprints to verify that they didn't change
+    let previousFingerprint = await trustStore.getFingerPrintFromRecord(subject);
+    let currentFingerprint = await this.getFingerPrint();
+    return previousFingerprint === currentFingerprint.toString();
```
mykola-mokhnach: I'd expect these both methods to return results of the same type

dpgraham:
Fixes pushed

dpgraham:
I went ahead and deleted the `uninstallSSLCert` method because we won't be deleting certs that we install any more because it's more performant to just leave the cert on a Simulator between tests instead of having to install and reset every time.

mykola-mokhnach:
> I went ahead and deleted the uninstallSSLCert method 

this will require to increase the major module version.
Also, this method might be useful while replacing the cert (if such feature is ever needed)

dpgraham:
@mykola-mokhnach We do currently replace the cert. Calling `installSSLCert` will replace what is already there.

@imurchie What do you think? Should we leave in `uninstallSSLCert` or take it out because it's unused.

PR #154, opened 2018-06-05
Add faceId biometrics

dpgraham:
* Went in to fix IOS Simulator because tests were failing for touchId on iPhone X (because iPhone X doesn't support touchId)
* Went ahead and added `faceId` as a biometric that can also be used alongside `touchId` since it's something that has been asked for and probably will be in high demand
* Updated tests to test both and also updated tests to catch exceptions
* Will update `appium-xcuitest-driver` next to use the new interface

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -891,7 +891,7 @@ class SimulatorXcode6 extends EventEmitter {
       } catch (err) {
         log.errorAndThrow(`Could not complete operation. Make sure Simulator UI is running and the parent Appium application (e. g. Appium.app or Terminal.app) ` +
                           `is present in System Preferences > Security & Privacy > Privacy > Accessibility list. If the operation is still unsuccessful then ` +
-                          `it is not supported by this Simulator` +
+                          `it is not supported by this Simulator. ` +
                           `Original error: ${err.message}`);
       }
     });
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -901,7 +901,7 @@ class SimulatorXcode6 extends EventEmitter {
    *
    * @return {boolean} True if Touch ID Enrollment menu item is checked in Simulator menu
    */
-  async isTouchIDEnrolled () {
+  async isBiometricEnrolled () {
     const output = await this.executeUIClientScript(`
       tell application "System Events"
         tell process "Simulator"
@@ -919,7 +919,7 @@ class SimulatorXcode6 extends EventEmitter {
    *
    * @param {boolean} isEnabled - Set it to false in order to uncheck 'Toggle Enrolled State' flag
    */
-  async enrollTouchID (isEnabled = true) {
+  async enrollBiometric (isEnabled = true) {
     await this.executeUIClientScript(`
       tell application "System Events"
         tell process "Simulator"
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -889,8 +889,9 @@ class SimulatorXcode6 extends EventEmitter {
         const {stdout} = await exec('osascript', ['-e', resultScript]);
         return stdout;
       } catch (err) {
-        log.errorAndThrow(`Make sure Simulator UI is running and the parent Appium application (e. g. Appium.app or Terminal.app) ` +
-                          `is present in System Preferences > Security & Privacy > Privacy > Accessibility list. ` +
+        log.errorAndThrow(`Could not complete operation. Make sure Simulator UI is running and the parent Appium application (e. g. Appium.app or Terminal.app) ` +
+                          `is present in System Preferences > Security & Privacy > Privacy > Accessibility list. If the operation is still unsuccessful then it` +
+                          `it is not supported by this Simulator` +
                           `Original error: ${err.message}`);
       }
     });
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-6.js
```diff
@@ -889,8 +889,9 @@ class SimulatorXcode6 extends EventEmitter {
         const {stdout} = await exec('osascript', ['-e', resultScript]);
         return stdout;
       } catch (err) {
-        log.errorAndThrow(`Make sure Simulator UI is running and the parent Appium application (e. g. Appium.app or Terminal.app) ` +
-                          `is present in System Preferences > Security & Privacy > Privacy > Accessibility list. ` +
+        log.errorAndThrow(`Could not complete operation. Make sure Simulator UI is running and the parent Appium application (e. g. Appium.app or Terminal.app) ` +
+                          `is present in System Preferences > Security & Privacy > Privacy > Accessibility list. If the operation is still unsuccessful then ` +
+                          `it is not supported by this Simulator` +
```
mykola-mokhnach: Missing space between `Simulator ` and `Original error`

lib/simulator-xcode-6.js
```diff
@@ -901,7 +902,7 @@ class SimulatorXcode6 extends EventEmitter {
    *
    * @return {boolean} True if Touch ID Enrollment menu item is checked in Simulator menu
    */
-  async isTouchIDEnrolled () {
+  async isBiometricEnrolled () {
```
mykola-mokhnach: I assume merging this PR will require to raise the major version of this lib

lib/simulator-xcode-6.js
```diff
@@ -919,7 +920,7 @@ class SimulatorXcode6 extends EventEmitter {
    *
    * @param {boolean} isEnabled - Set it to false in order to uncheck 'Toggle Enrolled State' flag
    */
-  async enrollTouchID (isEnabled = true) {
+  async enrollBiometric (isEnabled = true) {
     await this.executeUIClientScript(`
       tell application "System Events"
         tell process "Simulator"
```
mykola-mokhnach: Should also menu item names be updated in the script?

lib/simulator-xcode-9.js
```diff
@@ -272,31 +281,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
   }
 
   /**
-   * Get the current state of Touch ID Enrollment feature.
+   * Get the current state of a biometric (touch id, face id) enrollment.
    * @override
    */
-  async isTouchIDEnrolled () {
+  async isBiometricEnrolled (biometricName='touchId') {
```
mykola-mokhnach: missing space between the argument and the default value

mykola-mokhnach: the new parameter description should be in docstring

mykola-mokhnach: the default value should be `BIOMETRICS`property

lib/simulator-xcode-9.js
```diff
@@ -272,31 +281,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
     return _.isString(output) && output.trim() === 'true';
   }
 
   /**
-   * Execute a special Apple script, which changes Touch ID feature testing in Simulator UI client.
+   * Execute a special Apple script, which enrolls biometric (TouchId, FaceId) feature testing in Simulator UI client.
    * @override
    */
-  async enrollTouchID (isEnabled = true) {
+  async enrollBiometric (isEnabled = true, biometricName='touchId') {
```
mykola-mokhnach: same as above

lib/simulator-xcode-9.js
```diff
@@ -272,31 +281,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+    if (!BIOMETRICS[biometricName]) {
+      throw new Error(`${biometricName} is not a valid biometric. Use one of: ${JSON.stringify(_.keys(BIOMETRICS))}`);
+    }
+    const {menuName} = BIOMETRICS[biometricName];
```
mykola-mokhnach: I'd rather move this code block to a separate `verifyBiometricName` helper

lib/simulator-xcode-9.js
```diff
@@ -12,6 +12,15 @@ const SIMULATOR_SHUTDOWN_TIMEOUT = 15 * 1000;
 const startupLock = new AsyncLock();
 const preferencesPlistGuard = new AsyncLock();
 
+const BIOMETRICS = {
```
mykola-mokhnach: I expect this structure to be exported by this lib. You could make it a static class field

lib/simulator-xcode-6.js
dpgraham: Those are updated in the xcode-9 simulator.... for xcode-6 it just assumes touchId

lib/simulator-xcode-9.js
jlipps: to make this not a breaking change you could simply keep the `isTouchIDEnrolled` method and `enrollTouchID` method and have them call the more general one. not sure if we care.

dpgraham: I thought that too, but since appium-xcuitest-driver is the only dependent, I think bumping the major version should be okay.

lib/simulator-xcode-6.js
dpgraham: It will, yes.

PR #155, opened 2018-06-06
Add support for Xcode 10

imurchie:
This involves two things, so far:
* Make it so we can get the correct class
* Update Calendar access support to handle the added 3 fields to the table in the db


lib/simulator.js
```diff
@@ -36,27 +37,37 @@ async function getSimulator (udid) {
...
+        SimClass = SimulatorXcode9;
+      } else {
+        SimClass = SimulatorXcode93;
+      }
+      break;
+    case 10:
+      SimClass = SimulatorXcode10;
+      break;
+    default:
+      handleUnsupportedXcode(xcodeVersion);
```
mykola-mokhnach: Cannot we just default to the most recent simulator class by default instead of throwing an exception? I don't see any need in having derived classes, that are empty

lib/simulator.js
```diff
@@ -14,9 +14,6 @@ function handleUnsupportedXcode (xcodeVersion) {
     throw new Error(`Tried to use an iOS simulator with xcode ` +
                     `version ${xcodeVersion.versionString} but only Xcode version ` +
                     `6.0.0 and up are supported`);
-  } else if (xcodeVersion.major >= 10) {
-    throw new Error(`Xcode version ${xcodeVersion.versionString} is ` +
```
mykola-mokhnach: Can we print a warning message instead?

mykola-mokhnach: About the compatibility with Xcode version X has not been tested for this Appium version, so one might expect issues.

lib/simulator.js
```diff
@@ -36,27 +33,38 @@ async function getSimulator (udid) {
...
-    }
-  } else if (xcodeVersion.major === 8) {
-    return new SimulatorXcode8(udid, xcodeVersion);
-  } else if (xcodeVersion.major === 9) {
-    if (xcodeVersion.minor < 3) {
-      return new SimulatorXcode9(udid, xcodeVersion);
-    } else {
-      return new SimulatorXcode93(udid, xcodeVersion);
-    }
+  let SimClass;
```
mykola-mokhnach: it's a bit confusing to start local variable names with uppercase char

lib/simulator.js
imurchie: Xcode 5 will not work with Appium. So an exception is warranted.

lib/simulator.js
imurchie: Is it? It will be holding a class. It's also assigned within a couple of lines. I don't find it confusing at all.

lib/simulator.js
mykola-mokhnach: I'm talking about the case if Xcode version is greater than the last known one, but in case you'd like to keep it as is then I don't mind

imurchie: That exception is gone here. A new Xcode major version will just get the latest known version.

lib/simulator.js
jlipps: i agree with @imurchie on this one, it even has `Class` in the variable name which is what makes it not confusing to me.

PR #158, opened 2018-06-12
Set simulator window to be rotated automatically when asked

mykola-mokhnach:
We've finally got a response from Apple and they suggested to enable Hardware->Rotate Device Automatically Simulator option in order for its window to follow the actual screen orientation set on XCUIDevice instance.


mykola-mokhnach:
Yeah, not actually my, but our company developers, since we post all issues from one account. Perhaps, the fact we were talking to Apple guys during WWDC makes a difference.

PR #161, opened 2018-07-08
fix getting biometric

KazuCocoa:
I encountered below error when I called touch id enrollment feature.

> @driver.toggle_touch_id_enrollment true
Selenium::WebDriver::Error::UnknownError: An unknown server-side error occurred while processing the command. Original error: [object Object] is not a valid biometric. Use one of: ["touchId","faceId"]
from UnknownError: An unknown server-side error occurred while processing the command. Original error: [object Object] is not a valid biometric. Use one of: ["touchId","faceId"]

- `biometricName` is `{menuName: "Touch id"}`
- `BIOMETRICS` is `{touchId: {menuName: "Touch id"}, faceId: {menuName: "Face id"}}`
- Hence, `BIOMETRICS[biometricName]` never be true

~~This PR fix the behaviour to make sure the `menuName` exists in `BIOMETRICS`.~~
Use 'touchId' instead of `SimulatorXcode9.BIOMETRICS.touchId`

~~In JS world, is there a good way to get key by `biometricName`?~~
I mean if `biometricName` can be `touchId`, not `{menuName: "Touch id"}`,  `BIOMETRICS[biometricName]` works properly. In the case, we need to change https://github.com/KazuCocoa/appium-ios-simulator/blob/476635a6ec67609337468176fbff24ade7f1dcf3/lib/simulator-xcode-9.js#L304 and https://github.com/KazuCocoa/appium-ios-simulator/blob/476635a6ec67609337468176fbff24ade7f1dcf3/lib/simulator-xcode-9.js#L323 .

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -286,22 +286,24 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     `;
   }
 
+  /**
+   * Return proper biometric menu name object.
+   * @param {string} biometricName A key in BIOMETRICS like 'bundleId'
+   * @returns {object} Object in BIOMETRICS
+   */
   getBiometric (biometricName) {
     const {BIOMETRICS} = SimulatorXcode9;
-    const bio = Object.keys(BIOMETRICS).filter(function (key) {
-      return BIOMETRICS[key].menuName === biometricName.menuName;
-    });
-    if (bio.length === 0) {
+    if (!BIOMETRICS[biometricName]) {
       throw new Error(`${biometricName} is not a valid biometric. Use one of: ${JSON.stringify(_.keys(BIOMETRICS))}`);
     }
-    return biometricName;
+    return BIOMETRICS[biometricName];
   }
 
   /**
    * Get the current state of a biometric (touch id, face id) enrollment.
    * @override
    */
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-9.js
```diff
@@ -288,10 +288,13 @@ class SimulatorXcode9 extends SimulatorXcode8 {
 
   getBiometric (biometricName) {
     const {BIOMETRICS} = SimulatorXcode9;
-    if (!BIOMETRICS[biometricName]) {
+    const bio = Object.keys(BIOMETRICS).filter(function (key) {
```
mykola-mokhnach: `const bio = _.keys(BIOMETRICS).filter((key) => BIOMETRICS[key].menuName === biometricName)`

mykola-mokhnach:
I think the code there is correct. It is just the default value at 

`async isBiometricEnrolled (biometricName=SimulatorXcode9.BIOMETRICS.touchId) {`

is incorrect and should be replaced with just `biometricName = 'touchId'`

KazuCocoa:
I agree with the way.
I confirmed it worked correctly with 1fb1d53

KazuCocoa:
@imurchie Could you release new version to fix an error of toggle_touch_id_enrollment when you have a time?

imurchie:
Published in `appium-ios-simulator@2.14.1`.

PR #162, opened 2018-07-12
Fall back to non-AppleScript UI state detection if necessary permissions are missing

mykola-mokhnach:
It looks like Apple is getting super paranoid about security in Mojave-beta. This patch allows to avoid errors in UI state detection


mykola-mokhnach:
more info on the topic: https://www.felix-schwarz.org/blog/2018/06/apple-event-sandboxing-in-macos-mojave

imurchie:
Published in `appium-ios-simulator@2.14.2`.

PR #164, opened 2018-07-29
Add method to change ReduceMotion setting for simulators

vrunoa:
fyi @jlipps 

Add the possibility to set ReduceMotion in simulators, this can help reduce flakiness of tests.


lib/settings.js
```diff
@@ -132,6 +134,17 @@ async function updateLocationSettings (sim, bundleId, authorized) {
   }
 }
 
+async function reduceMotion (sim, reduceMotion = true) {
+  let paths = await plistPaths(sim, 'accesibilitySettings');
+  for (let i in paths) {
```
mykola-mokhnach: use `of` instead

lib/settings.js
```diff
@@ -132,6 +134,17 @@ async function updateLocationSettings (sim, bundleId, authorized) {
...
+    let file = paths[i];
+    let updates = {
+      ReduceMotionEnabled: reduceMotion ? 1 : 0
```
mykola-mokhnach: is it numeric or boolean parameter in the actual plist?

lib/settings.js
```diff
@@ -132,6 +134,17 @@ async function updateLocationSettings (sim, bundleId, authorized) {
...
+    };
+    await update(file, updates);
```
mykola-mokhnach: the intermediate `updates` variable is not needed

lib/simulator-xcode-6.js
```diff
@@ -617,6 +617,25 @@ class SimulatorXcode6 extends EventEmitter {
...
   }
 
+  /**
+   * Enable/Disable reduce motion.
+   *
+   * @param {string} bundleId - The application ID to update.
+   * @param {boolean} reduceMotion - Whether or not to enable it.
+   */
+  async reduceMotion (bundleID, reduceMotion = true) {
+    if (await this.isRunning()) {
```
mykola-mokhnach: I would not do daemon killing here, since it will an expected side effect

lib/simulator-xcode-6.js
```diff
@@ -617,6 +617,25 @@ class SimulatorXcode6 extends EventEmitter {
...
+      await this.endSimulatorDaemon();
+    } else if (await this.isFresh()) {
+      await this.launchAndQuit(false, STARTUP_TIMEOUT);
+    }
+    await settings.reduceMotion(this, bundleID, reduceMotion);
+  }
+
+  async readSettings (plist) {
```
mykola-mokhnach: what is the purpose of `readSettings` method?

test/functional/settings-e2e-specs.js
```diff
@@ -0,0 +1,55 @@
...
+    if (devicePresent) {
+      await killAllSimulators();
+      await simctl.deleteDevice(udid);
+    }
+  });
+
+  it("check accesibility reduce motion settings", async function () {
+    let sim = await getSimulator(udid);
+    await sim.reduceMotion(true);
+    let fileSettings = await sim.readSettings('accesibilitySettings');
```
mykola-mokhnach: it is, actually, spelled as `accessibility`

test/functional/settings-e2e-specs.js
```diff
@@ -0,0 +1,55 @@
...
+    for (let file in fileSettings) {
+      let settings = fileSettings[file];
```
mykola-mokhnach: use `of`

lib/settings.js
vrunoa: @mykola-mokhnach is an integer.

lib/simulator-xcode-6.js
vrunoa: @mykola-mokhnach the reduce motion change wont be available if we dont reboot the simulator. Let change that kill for reboot.

lib/simulator-xcode-6.js
vrunoa: I was using it for functional testing, let me see if I can have a different approach besides exposing this method

lib/simulator-xcode-6.js
```diff
@@ -612,6 +612,21 @@ class SimulatorXcode6 extends EventEmitter {
     return await settings.updateLocationSettings(this, bundleId, authorized);
   }
 
+  /**
+   * Enable/Disable reduce motion.
+   *
+   * @param {string} bundleId - The application ID to update.
+   * @param {boolean} reduceMotion - Whether or not to enable it.
+   */
+  async reduceMotion (reduceMotion = true) {
```
mykola-mokhnach: I'd rather rename the method to setReduceMotion

lib/simulator-xcode-6.js
```diff
@@ -612,6 +612,21 @@ class SimulatorXcode6 extends EventEmitter {
...
+    if (await this.isRunning()) {
```
mykola-mokhnach: I still consider that restart should not be in this method. It should be the caller code, who is responsible for simulator reboot (and it has to be stated in the docstring, that the method has no effect unless the sim is rebooted if it is currently running)

jlipps: Yeah, I agree. We have (or used to have) code in appium-ios-driver / appium-xcuitest-driver that check the capabilities a user sends in to determine if any of them will require a sim reboot, and it handles the reboot after running the commands from this package. It should be up to the caller to do the reboot.

test/functional/settings-e2e-specs.js
```diff
@@ -0,0 +1,60 @@
...
+      }).length > 0;
+    if (devicePresent) {
+      await killAllSimulators();
+      await simctl.deleteDevice(udid);
+    }
+  });
+
+  it("check accesibility reduce motion settings", async function () {
+    let sim = await getSimulator(udid);
+    await sim.reduceMotion(true);
```
imurchie: `setReduceMotion`?

vrunoa: thanks

test/unit/settings-specs.js
```diff
@@ -270,4 +269,31 @@ describe('settings', function () {
...
+      sim = new SimulatorXcode6();
+      sinon.stub(settings, 'setReduceMotion');
+    });
+
+    afterEach(async function () {
+      sinon.restore();
+      sinon.stub(settings, 'setReduceMotion');
+    });
+
+    it('should launch simulator if not fresh before setting reduce motion', async function () {
```
mykola-mokhnach: perhaps `if fresh`

PR #165, opened 2018-08-06
Update for Xcode 10b5

imurchie:
Xcode 10b5 introduced another field to the access database, as well as the situation where the app container has a value, but it does not exist yet.

### DIFF
--- a/lib/calendar.js
+++ b/lib/calendar.js
@@ -60,7 +60,7 @@ export default class Calendar {
     } else {
       query = this.xcodeVersion.major < 10 ?
         `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0);` :
-        `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0, 0, 0, 0);`;
+        `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0, 0, 0, 0, 0);`;
     }
     await execSQLiteQuery(db, query, bundleID);
   }
--- a/lib/simulator-xcode-10.js
+++ b/lib/simulator-xcode-10.js
@@ -1,0 +1,36 @@
+import SimulatorXcode93 from './simulator-xcode-9.3';
+import { getAppContainer, appInfo } from 'node-simctl';
+import { fs } from 'appium-support';
+
+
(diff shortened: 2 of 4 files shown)


lib/simulator-xcode-10.js
```diff
@@ -0,0 +1,36 @@
...
+      const appContainer = await getAppContainer(this.udid, bundleId, false);
+      const exists = await fs.exists(appContainer);
+      return exists && appContainer.endsWith('.app');
+    } catch (err) {
+      // get_app_container subcommand fails for system applications,
+      // so we try the hidden appinfo subcommand, which prints correct info for
+      // system/hidden apps
+      try {
+        const info = await appInfo(this.udid, bundleId);
+        return info.indexOf('ApplicationType') !== -1;
```
mykola-mokhnach: includes?

lib/simulator-xcode-10.js
```diff
@@ -0,0 +1,36 @@
...
+   * @override
+   *
+   * @param {string} bundleId - The bundle id of the application to be checked.
+   * @return {boolean} True if the given application is installed.
+   */
+  async isAppInstalled (bundleId) {
+    try {
+      const appContainer = await getAppContainer(this.udid, bundleId, false);
+      const exists = await fs.exists(appContainer);
+      return exists && appContainer.endsWith('.app');
```
mykola-mokhnach: should we use short calculation here (if appContainer.endsWith('.app') is false then there is no point to check existence)

lib/calendar.js
```diff
@@ -60,7 +60,7 @@ export default class Calendar {
     } else {
       query = this.xcodeVersion.major < 10 ?
         `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0);` :
-        `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0, 0, 0, 0);`;
+        `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0, 0, 0, 0, 0);`;
```
mykola-mokhnach: is this safe for previous DB versions?

imurchie: I don't think that it matters, for betas. There is no reason to support a previous beta version of Xcode.

lib/simulator-xcode-10.js
imurchie: Good idea.

PR #166, opened 2018-08-09
Only update settings if necessary

imurchie:
We have all the machinery in place to check if requested settings values are different from current values. This can be important since changed values means rebooting a simulator. Calling code will have to be updated to take the return value and either reboot or not, depending on need.

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -61,27 +61,32 @@ async function plistPaths (sim, identifier) {
 }
 
 async function updateSettings (sim, plist, updates) {
-  let paths = await plistPaths(sim, plist);
-  for (let path of paths) {
-    await update(path, updates);
+  let updated = false;
+  for (let path of await plistPaths(sim, plist)) {
+    updated = await update(path, updates) || updated;
   }
+  return updated;
 }
 
 // update a plist file, located at pathToPlist
 // pass in an object, all settings specified in the object will be
 // updated on the plist, all others left as-is
 async function update (pathToPlist, updates) {
-  let currentSettings = await read(pathToPlist);
-  let newSettings = _.merge(currentSettings, updates);
-  await plist.updatePlistFile(pathToPlist, newSettings, true, false);
+  const currentSettings = await read(pathToPlist);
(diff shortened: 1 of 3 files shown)


lib/settings.js
```diff
@@ -155,25 +164,29 @@ async function updateSafariUserSettings (sim, settingSet) {
...
-      if (!_.has(userSettingSet, 'restrictedBool')) {
-        userSettingSet.restrictedBool = {};
-      }
-      for (let [key, value] of _.toPairs(newUserSettings)) {
-        userSettingSet.restrictedBool[key] = {value};
-      }
 
-      // actually do the update
-      await update(file, userSettingSet);
+  if (_.size(newUserSettings) === 0) {
```
mykola-mokhnach: isEmpty might look better

lib/simulator-xcode-6.js
```diff
@@ -630,8 +630,9 @@ class SimulatorXcode6 extends EventEmitter {
    * @param {object} updates - The hash of key/value pairs to update for Safari.
    */
   async updateSafariSettings (updates) {
-    await settings.updateSafariUserSettings(this, updates);
-    await settings.updateSettings(this, 'mobileSafari', updates);
+    let updated = await settings.updateSafariUserSettings(this, updates);
+    updated = await settings.updateSettings(this, 'mobileSafari', updates) || updated;
+    return updated;
```
mykola-mokhnach: perhaps `return await settings.updateSafariUserSettings(this, updates) | await settings.updateSettings(this, 'mobileSafari', updates)` ?

lib/settings.js
```diff
@@ -60,27 +60,32 @@ async function plistPaths (sim, identifier) {
 }
 
 async function updateSettings (sim, plist, updates) {
-  let paths = await plistPaths(sim, plist);
-  for (let path of paths) {
-    await update(path, updates);
+  let updated = false;
```
mykola-mokhnach: this could also use B.reduce

imurchie: I **do** like a reason to use `reduce`.

PR #167, opened 2018-08-11
Dynamically figure out column count for calendar query

imurchie:
The query is all over the place, but we just need it to be filled with `0`s.  So query the number of columns and fill that number in.

### DIFF
--- a/lib/calendar.js
+++ b/lib/calendar.js
@@ -1,6 +1,9 @@
 import path from 'path';
 import { fs, mkdirp } from 'appium-support';
 import { execSQLiteQuery } from './utils';
+import _ from 'lodash';
+import log from './logger';
+
 
 export default class Calendar {
 
@@ -14,7 +17,7 @@ export default class Calendar {
       return this.db;
     }
 
-    let tccPath = path.resolve(this.sharedResourcesDir, 'Library', 'TCC');
+    const tccPath = path.resolve(this.sharedResourcesDir, 'Library', 'TCC');
     if (!(await fs.exists(tccPath))) {
       await mkdirp(tccPath);
     }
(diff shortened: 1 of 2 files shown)


lib/calendar.js
```diff
@@ -58,15 +61,27 @@ export default class Calendar {
         csreq=0
       WHERE client='?' AND service='kTCCServiceCalendar'`.replace(/\n/g, ' ');
     } else {
-      query = this.xcodeVersion.major < 10 ?
-        `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0);` :
-        `INSERT INTO 'access' VALUES ('kTCCServiceCalendar', '?', 0, 1, 1, 0, 0, 0, 0, 0, 0);`;
+      let count = this.columnCount;
+      if (_.isEmpty(this.columnCount)) {
```
mykola-mokhnach: note, that isEmpty returns true for primitive types, like numbers

lib/calendar.js
```diff
@@ -58,15 +61,27 @@ export default class Calendar {
...
+        try {
+          const {stdout} = await execSQLiteQuery(db, `pragma table_info('access')`);
+          // stdout has a number of records, each starting with `cid`
+          count = this.columnCount = stdout.trim().split('cid').filter((record) => !_.isEmpty(record)).length;
```
mykola-mokhnach: I think it might be easier to read this line if formatted like

```
stdout.trim()
  .split
  .filter
...
```

lib/calendar.js
imurchie: That was the intention. Though I guess `_.isNumber` would better signal the intent.

lib/calendar.js
imurchie: Sure.

PR #169, opened 2018-08-12
Tune isUIClientRunning verification

mykola-mokhnach:
Mac OS Mojave blocks Apple script execution without special permissions, so we must have a fallback there.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -85,7 +85,7 @@ class SimulatorXcode6 extends EventEmitter {
    *
    * @return {?string} The process ID or null if the UI client is not running
    */
-  async uiClientPid () {
+  async getUIClientPid () {
     let stdout;
     try {
       ({stdout} = await exec('osascript', [
@@ -100,11 +100,12 @@ class SimulatorXcode6 extends EventEmitter {
         return null;
       }
     }
-    if (!isNaN(parseInt(stdout, 10))) {
-      log.debug(`Retrieved Simulator's UI client PID: ${stdout.trim()}`);
-      return stdout.trim();
+    if (isNaN(parseInt(stdout, 10))) {
+      return null;
     }
-    return null;
+    stdout = stdout.trim();
+    log.debug(`Got Simulator UI client PID: ${stdout}`);
+    return stdout;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-6.js
```diff
@@ -72,18 +73,48 @@ class SimulatorXcode6 extends EventEmitter {
...
+   * Retrieves the current process id of the UI client
+   *
+   * @return {?string} The process ID or null if the UI client is not running
+   */
+  async getUIClientPid () {
+    let stdout;
+    try {
+      ({stdout} = await exec('osascript', [
+        '-e',
+        `tell application "System Events" to unix id of processes whose bundle identifier is "${this.uiClientBundleId}"`
```
KazuCocoa: Do you know the history of apple script and pgrep here?
@mykola-mokhnach @imurchie 

When I ran below command and compared the result, pgrep was much faster than AppleScript.
Rarelly AppleScript hanged(?) and returned the result a couple of minutes later. 

```
$ time osascript -e 'tell application "System Events" to unix id of processes whose bundle identifier is "com.apple.iphonesimulator"'
29513

real    0m1.192s
user    0m0.132s
sys     0m0.114s

$ time pgrep -xni 'Simulator'
29513

real    0m0.035s
user    0m0.005s
sys     0m0.006s
```

I was not familiar with the history here.
I just thought the process can be faster if we could use `pgrep` first and fallback to the apple script.


(I know the apple script and pgrep logic was not introduced in this PR though.)

mykola-mokhnach: I used Apple Script to make sure we get the correct application id(s) in case multiple processes with the same name are running. AS is able to uniquely connect an application bundle to process id and pgrep works on the lower level. Although, I don't mind if we only use pgrep if we don't really care about the presence of processes with similar names for he sake of better performance.

KazuCocoa: I see.
thanks!

NozomiIto: How about using `ps` command etc?
It can get the full path of Simulator app like `/Applications/Xcode.app/Contents/Developer/Applications/Simulator.app/Contents/MacOS/Simulator` and almost surely can identify if Simulator.app is launched. And maybe more stable (and faster ) than Apple script.

KazuCocoa: a note: are `-f` and a path in `Xcode.app` much reliable?

```
$ time pgrep -f ‘Contents/Developer/Applications/Simulator.app’
50404

real    0m0.037s
user    0m0.008s
sys    0m0.004s
```

mykola-mokhnach: yes, this makes more sense. Will change it soon

PR #173, opened 2018-09-17
Increase retries for Xcode 10 sim starting

imurchie:
New sims on Xcode 10 take a while.

Also add Xcode 10 to the Travis build.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -569,13 +569,17 @@ class SimulatorXcode6 extends EventEmitter {
     // locally, 7.1 averages 8.5 retries (from 6 - 12)
     //          8 averages 0.6 retries (from 0 - 2)
     //          9 averages 14 retries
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-6.js
```diff
@@ -569,13 +569,17 @@ class SimulatorXcode6 extends EventEmitter {
...
+    try {
+      await retryInterval(60, 250, async () => {
+        if (await this.isFresh()) {
+          let msg = 'Simulator files not fully created. Waiting a bit';
+          log.debug(msg);
+          throw new Error(msg);
+        }
+      });
+    } catch (err) {
+      log.warn(`Timeout waiting for simulator files to be created. Continuing`);
```
jlipps: it's possible that it could work even if we hit the timeout?

imurchie: Yes. That's why I added the catch and log. Previously there was an unhandled promise rejection.

PR #177, opened 2018-10-01
Update appium-gulp-plugins

imurchie:
This PR does two main things:
1. Updates `eslint-config-appium` and fixes any linting errors.
1. Updated `appium-gulp-plugins` and moves to Babel 7 and Gulp 4.

See https://github.com/appium/appium-gulp-plugins/pull/35 for more details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 os: osx
 language: node_js
-node_js: "6"
+node_js: "8"
 matrix:
   include:
     - osx_image: xcode7.3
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: required
 os: osx
 language: node_js
-node_js: "6"
+node_js: "8"
```
mykola-mokhnach: why not 10?

imurchie: Because 8 is the minimum.

PR #183, opened 2018-11-02
Implement more reliable flow for biometric

mykola-mokhnach:
The original idea is borrowed from https://github.com/wix/AppleSimulatorUtils/blob/master/applesimutils/applesimutils/main.m


lib/simulator-xcode-9.js
```diff
@@ -291,55 +282,61 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+   * @override
    */
-  getBiometric (biometricName) {
-    const {BIOMETRICS} = SimulatorXcode9;
-    if (!BIOMETRICS[biometricName]) {
-      throw new Error(`${biometricName} is not a valid biometric. Use one of: ${JSON.stringify(_.keys(BIOMETRICS))}`);
+  async isBiometricEnrolled () {
+    const {stdout} = await spawn(this.udid, [
+      'notifyutil',
+      '-g', ENROLLMENT_NOTIFICATION_RECEIVER
```
KazuCocoa: [nits]
What about adding `-v` to show a log with timestamp, key, etc?

lib/simulator-xcode-8.js
```diff
@@ -210,13 +210,61 @@ class SimulatorXcode8 extends SimulatorXcode7 {
...
+
+  /**
+   * @inheritdoc
+   * @override
+   */
+  async isBiometricEnrolled () {
+    const output = await this.executeUIClientScript(`
+      tell application "System Events"
+        tell process "Simulator"
+          set dstMenuItem to menu item "Toggle Enrolled State" of menu 1 of menu item "Touch ID" of menu 1 of menu bar item "Hardware" of menu bar 1
```
dpgraham: ~~Are we able to do the same thing for `Face Id`?~~

dpgraham: Nvm, just remembered that `face id` doesn't apply to this xcode version.

lib/simulator-xcode-9.js
```diff
@@ -291,55 +282,61 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
-        end tell
-      end tell
-    `);
+  async sendBiometricMatch (shouldMatch = true, biometricName = 'touchId') {
+    const domainComponent = toBiometricDomainComponent(biometricName);
+    const domain = `com.apple.BiometricKit_Sim.${domainComponent}.${shouldMatch ? '' : 'no'}match`;
+    await spawn(this.udid, [
+      'notifyutil',
+      '-p', domain
+    ]);
```
dpgraham: Good find. This makes me feel a lot better than using AppleScript.

PR #198, opened 2019-02-05
Add ability to change log prefix depending on platform

imurchie:
Updating logging so the prefix is changed depending on the platform. This will not work if there are multiple simulators of different platforms (since the logger is shared) but that bridge can be crossed if/when it is reached.

Logs become like:
```
info tvOSSim Constructing tvOS simulator for Xcode version 9.4.1 with udid 'C2DEBF4F-D7D4-4A3F-8A31-13E86B505B1B'
info iOSSim Constructing iOS simulator for Xcode version 9.4.1 with udid '697D0C48-A7E6-4BBC-9112-015BE58D9432'
info watchOSSim Constructing watchOS simulator for Xcode version 9.4.1 with udid '861897A6-9E93-4423-9D5E-8D9FCDD84611'
```

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -43,6 +43,7 @@ class SimulatorXcode6 extends EventEmitter {
    */
   constructor (udid, xcodeVersion) {
     super();
+
     this.udid = String(udid);
     this.xcodeVersion = xcodeVersion;
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -162,14 +162,10 @@ async function getSimulatorInfo (udid) {
   // see the README for github.com/appium/node-simctl for example output of getDevices()
   let devices = await getDevices();
 
-  devices = _.toPairs(devices).map((pair) => {
-    return pair[1];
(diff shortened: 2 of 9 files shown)


lib/simulator-xcode-6.js
```diff
@@ -43,6 +43,7 @@ class SimulatorXcode6 extends EventEmitter {
    */
   constructor (udid, xcodeVersion) {
     super();
+
```
KazuCocoa: new empty line?

lib/utils.js
```diff
@@ -167,11 +167,15 @@ async function simExists (udid) {
   }).reduce((a, b) => {
     return a.concat(b);
```
mykola-mokhnach: these returns are redundant

mykola-mokhnach: reduce can be replaced with flatMap

imurchie: I'd rather not have to mess with polyfills to get functionality we can already get from existing functions.

mykola-mokhnach: I mean lodash's flatMap

PR #227, opened 2019-06-17
Update headless logic

mykola-mokhnach:
Addresses https://github.com/appium/appium/issues/12765

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -86,43 +86,52 @@ class SimulatorXcode9 extends SimulatorXcode8 {
         log.warn(`'xcrun simctl boot ${this.udid}' command has returned non-zero code. The problem was: ${err.stderr}`);
       }
     };
-    const waitForShutdown = async () => {
+    const waitForShutdown = async (waitMs = SIMULATOR_SHUTDOWN_TIMEOUT) => {
       try {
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-9.js
```diff
@@ -86,43 +86,52 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
-        }, {waitMs: SIMULATOR_SHUTDOWN_TIMEOUT, intervalMs: 500});
+        }, {waitMs, intervalMs: 500});
       } catch (err) {
-        throw new Error(`Simulator is not in 'Shutdown' state after ${SIMULATOR_SHUTDOWN_TIMEOUT}ms`);
+        throw new Error(`Simulator is not in 'Shutdown' state after ${waitMs}ms`);
       }
     };
     let shouldWaitForBoot = true;
     const startTime = process.hrtime();
     await startupLock.acquire(this.uiClientBundleId, async () => {
```
imurchie: `acquire` returns whatever the callback returns, so this can become `const shouldWaitForBoot = await startupLock.acquire(this.uiClientBundleId, async () => {` and then return a boolean rather than setting the boolean and returning.

mykola-mokhnach: thanks, changed

PR #234, opened 2019-08-05
Fix geolocation coordinates format

lamkovod:
Number formatting depends on System preferences and Locale.
Issues occurs if user is using commas as decimal separator.
As string formatting should fix this problem.


jlipps:
published in 3.12.2

PR #235, opened 2019-08-12
fix: get tests to run on xcode 11 in Travis

imurchie:
This started as just a simple addition of Xcode 11 into the build, but that failed. So...

1. Override `getLaunchDaemonsRoot` for Xcode 11 since the directory structure has changed from `[...]/Platforms/iPhoneOS.platform/Developer/Library/CoreSimulator/[...]` to `[...]/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/[...]`. _Thanks Apple._
1. Run permissions tests in Travis, since they did not work anywhere. This involved installing the Wix tooling, and adding a function to set a single permission (which seemed to be the idea behind what was written in the test). I also inverted the arguments to some functions so they are all the same.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -1125,6 +1125,11 @@ class SimulatorXcode6 extends EventEmitter {
 
   //endregion
 
+
+  async setPermission (bundleId, permission, value) {
+    await this.setPermissions(bundleId, {[permission]: value});
+  }
+
   /**
    * Sets the permissions for the particular application bundle.
    *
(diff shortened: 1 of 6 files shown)


lib/simulator-xcode-6.js
```diff
@@ -1125,6 +1125,11 @@ class SimulatorXcode6 extends EventEmitter {
 
   //endregion
 
+
+  async setPermission (bundleId, permission, value) {
```
mykola-mokhnach: please add docstring

PR #240, opened 2019-08-27
feat: add function to set global (device-level) safari settings

imurchie:
This can speed things up since the sim does not need to have been run, nor does Safari need to have loaded, to set properties on it.

This is a step toward getting https://github.com/appium/appium/issues/13069 to work.


mykola-mokhnach:
Can we have a test for this method?

PR #242, opened 2019-08-30
feat: get user installed bundleids by bundlename

KazuCocoa:
https://github.com/appium/appium-xcuitest-driver/pull/1052#discussion_r319355030

In like below path has user installed `.app`. This method read Info.plist in the `.app` and returns a list of bundle ids which has provide bundle name.
```
/Users/kazu/Library/Developer/CoreSimulator/Devices/E9245A58-B499-4B76-8FB1-7FA1342F3085/data/Containers/Bundle/Application/210A3413-B313-443A-A5EC-CB6CF1C5E84F/
```

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -188,12 +188,18 @@ class SimulatorXcode6 extends EventEmitter {
   /**
    * Returns user installed bundle ids which has 'bundleName' in their Info.Plist as 'CFBundleName'
    * @param {string} bundleId - The bundle id of the application to be checked.
+   * @return {array<string>} - The list of bundle ids which have 'bundleName'
    */
   async getUserInstalledBundleIdsByBundleName (bundleName) {
-    const rootUserAppDir = await this.buildBundlePathMap('Bundle');
+    const rootUserAppDir = await this.buildBundlePathMap('Bundl');
     const bundleIds = [];
     for (const [key, value] of Object.entries(rootUserAppDir)) {
       const infoPlist = await plist.parsePlistFile(
+        // var path = require('path');
+        // var EXTENSION = '.txt';
+        // var targetFiles = files.filter(function(file) {
+        //     return path.extname(file).toLowerCase() === EXTENSION;
+        // });
         path.resolve(`${await fs.glob(`${value}/*.app`)}`, 'Info.plist'), false);
       if (infoPlist.CFBundleName === bundleName) {
         bundleIds.push(key);
@@ -261,6 +267,11 @@ class SimulatorXcode6 extends EventEmitter {
         let bundleId = await readBundleId(dir);
         return {path: dir, bundleId};
       };
+    }
+
+    if (fs.exists(applicationList)) {
+      log.warn(`No directory path '${applicationList}'`);
+      return {};
     }
 
     let bundlePathDirs = await fs.readdir(applicationList);
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -210,7 +210,7 @@ class SimulatorXcode6 extends EventEmitter {
           bundleIds.push(bundleId);
         }
       } catch (err) {
-        log.warn(`Failed to read plist ${infoPlistPath}. Original error" '${err.message}'`);
+        log.warn(`Failed to read plist ${infoPlistPath}. Original error '${err.message}'`);
         continue;
       }
     }
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -2,7 +2,7 @@ import path from 'path';
 import * as simctl from 'node-simctl';
 import { default as xcode, getPath as getXcodePath } from 'appium-xcode';
 import log from './logger';
-import { fs, tempDir, mkdirp } from 'appium-support';
+import { fs, tempDir, mkdirp, plist } from 'appium-support';
 import B from 'bluebird';
 import _ from 'lodash';
 import AsyncLock from 'async-lock';
@@ -15,7 +15,6 @@ import extensions from './extensions/index';
 import { EventEmitter } from 'events';
 import Calendar from './calendar';
 import Permissions from './permissions';
-import { plist } from 'appium-support';
 
 
 const STARTUP_TIMEOUT = 60 * 1000;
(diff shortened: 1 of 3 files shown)


lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,24 @@ class SimulatorXcode6 extends EventEmitter {
     return appDirs.length !== 0;
   }
 
+  /**
+   * Returns user installed bundle ids which has 'bundleName' in their Info.Plist as 'CFBundleName'
+   * @param {string} bundleId - The bundle id of the application to be checked.
```
mykola-mokhnach: missing return param description

lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,24 @@ class SimulatorXcode6 extends EventEmitter {
...
+   */
+  async getUserInstalledBundleIdsByBundleName (bundleName) {
+    const rootUserAppDir = await this.buildBundlePathMap('Bundle');
+    const bundleIds = [];
+    for (const [key, value] of Object.entries(rootUserAppDir)) {
+      const infoPlist = await plist.parsePlistFile(
+        path.resolve(`${await fs.glob(`${value}/*.app`)}`, 'Info.plist'), false);
```
mykola-mokhnach: does fs.glob always result into something?

lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,24 @@ class SimulatorXcode6 extends EventEmitter {
     return appDirs.length !== 0;
   }
 
+  /**
+   * Returns user installed bundle ids which has 'bundleName' in their Info.Plist as 'CFBundleName'
+   * @param {string} bundleId - The bundle id of the application to be checked.
+   */
+  async getUserInstalledBundleIdsByBundleName (bundleName) {
+    const rootUserAppDir = await this.buildBundlePathMap('Bundle');
```
mykola-mokhnach: is the var always defined?

lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,24 @@ class SimulatorXcode6 extends EventEmitter {
     return appDirs.length !== 0;
   }
 
+  /**
+   * Returns user installed bundle ids which has 'bundleName' in their Info.Plist as 'CFBundleName'
+   * @param {string} bundleId - The bundle id of the application to be checked.
+   */
+  async getUserInstalledBundleIdsByBundleName (bundleName) {
```
mykola-mokhnach: can we have a test for this method?

lib/simulator-xcode-6.js
KazuCocoa: Yes. I checked some simulator instances. They had below structure when they had apps installed by us.
```
.../Bundle/Application/210A3413-B313-443A-A5EC-CB6CF1C5E84F/IntegrationApp.app/Info.plist (etc)
.../Bundle/Application/210A3413-B313-443A-A5EC-CB6CF1C5E84F/BundleMetadata.plist
```

imurchie:
On a sim that exists but has never been run, this causes:
```
dbug iOSSim Building bundle path map
(node:9664) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, scandir '/Users/isaac/Library/Developer/CoreSimulator/Devices/58E187EA-39D9-40A6-87FE-9B17BCEB5304/data/Containers/Bundle/Application'
(node:9664) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:9664) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,24 @@ class SimulatorXcode6 extends EventEmitter {
...
+    const rootUserAppDir = await this.buildBundlePathMap('Bundle');
+    const bundleIds = [];
+    for (const [key, value] of Object.entries(rootUserAppDir)) {
```
imurchie: Could better be `const [bundleId, filePath]` or something.

lib/simulator-xcode-6.js
KazuCocoa: Yes, I'm adding a case in simulator-e2e-specs

KazuCocoa:
ah, thanks

lib/simulator-xcode-6.js
imurchie: Not a big fan of using the default stringification of a single-entry array. It works for now, but it may not hold that only one `.app` will exist in the directory.

KazuCocoa:
A simulator has no `Containers` before launching it. So, maybe need to handle the case.

test/functional/simulator-e2e-specs.js
```diff
@@ -530,6 +530,41 @@ function runTests (deviceType) {
       });
     });
   });
+
+  describe('getUserInstalledBundleIdsByBundleName', function () {
```
imurchie: Does this _need_ a new sim? If not, it would be better to put it in the 'advanced features' section (https://github.com/appium/appium-ios-simulator/blob/master/test/functional/simulator-e2e-specs.js#L336). Otherwise this suite just gets longer and longer (especially on iOS 13, where sim startup is ridiculously slow).

lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,34 @@ class SimulatorXcode6 extends EventEmitter {
...
+  async getUserInstalledBundleIdsByBundleName (bundleName) {
+    const rootUserAppDir = await this.buildBundlePathMap('Bundle');
+    const bundleIds = [];
+    if (_.isEmpty(rootUserAppDir)) {
+      return bundleIds;
+    }
+
+    for (const [bundleId, userAppDirPath] of Object.entries(rootUserAppDir)) {
+      const appFile = (await fs.readdir(userAppDirPath)).find(
+        (file) => path.extname(file).toLowerCase() === '.app');
```
mykola-mokhnach: I personally find it more readable if the starting point is moved to the next line, for example

```
const r = arr.map(...)
  .filter(...)
  .find(...)
```

lib/simulator-xcode-6.js
```diff
@@ -185,6 +185,34 @@ class SimulatorXcode6 extends EventEmitter {
     return appDirs.length !== 0;
   }
 
+  /**
+   * Returns user installed bundle ids which has 'bundleName' in their Info.Plist as 'CFBundleName'
+   * @param {string} bundleId - The bundle id of the application to be checked.
+   * @return {array<string>} - The list of bundle ids which have 'bundleName'
+   */
+  async getUserInstalledBundleIdsByBundleName (bundleName) {
```
mykola-mokhnach: the name of this method and parameter types should match with the one we use for the real device in XCUITest driver, so we could polymorph them

test/functional/simulator-e2e-specs.js
KazuCocoa: Yeah, agree.
I need a simulator created but before launching since we could not find a case, `buildBundlePathMap` raise an error, in current test cases. Let me add the case in existing place.

PR #245, opened 2019-09-27
feat: add function to set dark mode in sim

imurchie:
Quite simple to set the dark mode ui preferences, and they can be done _before_ the sim is started!

Cc @wswebcreation

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -157,12 +157,12 @@ async function setReduceMotion (sim, reduceMotion = true) {
   }
 }
 
-async function setDarkMode (sim, darkMode = true) {
-  log.debug(`Setting dark mode ${darkMode ? 'on' : 'off'}`);
+async function setDarkMode (sim, isEnabled = true) {
+  log.debug(`Setting dark mode ${isEnabled ? 'on' : 'off'}`);
   const paths = await plistPaths(sim, 'uiStyleSettings');
   for (const file of paths) {
     await update(file, {
-      UserInterfaceStyleMode: darkMode ? 2 : 1,
+      UserInterfaceStyleMode: isEnabled ? 2 : 1,
     });
   }
 }
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -59,6 +59,9 @@ async function plistPaths (sim, identifier) {
     case 'accessibilitySettings':
       paths.push(path.resolve(simDirectory, 'Library', 'Preferences', 'com.apple.Accessibility.plist'));
       break;
+    case 'uiStyleSettings':
+      paths.push(path.resolve(simDirectory, 'Library', 'Preferences', 'com.apple.uikitservices.userInterfaceStyleMode.plist'));
+      break;
   }
 
   return paths;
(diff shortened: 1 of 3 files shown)


wswebcreation:
Wow, hero!

lib/settings.js
```diff
@@ -154,6 +157,16 @@ async function setReduceMotion (sim, reduceMotion = true) {
   }
 }
 
+async function setDarkMode (sim, darkMode = true) {
+  log.debug(`Setting dark mode ${darkMode ? 'on' : 'off'}`);
+  const paths = await plistPaths(sim, 'uiStyleSettings');
+  for (const file of paths) {
+    await update(file, {
```
mykola-mokhnach: This could be done in parallel using B.all

lib/simulator-xcode-6.js
```diff
@@ -698,6 +698,16 @@ class SimulatorXcode6 extends EventEmitter {
     await settings.setReduceMotion(this, reduceMotion);
   }
 
+  /**
+   * Enable/Disable dark mode. This only works on Xcode 11+/iOS 13+.
+   * This function can be called before the simulator has been run.
+   *
+   * @param {boolean} darakMode - Whether or not to enable it (true by default)
```
mykola-mokhnach: darakMode?

lib/simulator-xcode-11.js
```diff
@@ -16,6 +17,13 @@ class SimulatorXcode11 extends SimulatorXcode10 {
     return path.resolve(devRoot,
       'Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/LaunchDaemons');
   }
+
+  /**
+   * @override
+   */
+  async setDarkMode (darkMode = true) {
```
mykola-mokhnach: I'd rather change this to `setUiTheme (isDark = true)` to avoid unnecessary repetitions
or setDarkMode (isEnabled)

lib/settings.js
```diff
@@ -157,12 +157,12 @@ async function setReduceMotion (sim, reduceMotion = true) {
...
 
-async function setDarkMode (sim, darkMode = true) {
-  log.debug(`Setting dark mode ${darkMode ? 'on' : 'off'}`);
+async function setDarkMode (sim, isEnabled = true) {
+  log.debug(`Setting dark mode ${isEnabled ? 'on' : 'off'}`);
   const paths = await plistPaths(sim, 'uiStyleSettings');
   for (const file of paths) {
     await update(file, {
-      UserInterfaceStyleMode: darkMode ? 2 : 1,
+      UserInterfaceStyleMode: isEnabled ? 2 : 1,
```
mykola-mokhnach: If these are set by integer than it means Apple could add more style modes in the future. Do we still want to keep the current approach or make it more universal?

imurchie: I'd rather keep it as it is and deal with Apple's additions later, than try to predict what they might do.

PR #255, opened 2020-01-02
feat: allow passing in of platform

imurchie:
There are times when we know, when calling `getSimulator`, that the simulator exists (for instance, when we have just created it). On my machine, using iOS 13+, it takes around 10 seconds to check for the existence of the simulator, so skipping it is not a trivial thing.

### DIFF
--- a/lib/simulator.js
+++ b/lib/simulator.js
@@ -24,7 +24,7 @@ function handleUnsupportedXcode (xcodeVersion) {
  *
  * @param {string} udid - The ID of an existing Simulator.
  * @param {Object} opts - an Options object that can include the `platform` of
- *                        the simulator associated with the udid, and `checkExistance`,
+ *                        the simulator associated with the udid, and `checkExistence`,
  *                        which allows one to skip checking that the udid is valid
  *                        (defaults to `true`)
  * @throws {Error} If the Simulator with given udid does not exist in devices list.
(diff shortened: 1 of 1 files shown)


lib/simulator.js
```diff
@@ -23,23 +23,36 @@ function handleUnsupportedXcode (xcodeVersion) {
...
  * @throws {Error} If the Simulator with given udid does not exist in devices list.
  *   If you want to create a new simulator, you can use the `createDevice()` method of
  *   [node-simctl](github.com/appium/node-simctl).
  * @return {object} Simulator object associated with the udid passed in.
  */
-async function getSimulator (udid) {
+async function getSimulator (udid, opts = {}) {
+  let {
+    platform,
+    checkExistance = true,
```
mykola-mokhnach: checkExistence

lib/simulator.js
```diff
@@ -23,23 +23,36 @@ function handleUnsupportedXcode (xcodeVersion) {
...
+ *                        (defaults to `true`)
  * @throws {Error} If the Simulator with given udid does not exist in devices list.
  *   If you want to create a new simulator, you can use the `createDevice()` method of
  *   [node-simctl](github.com/appium/node-simctl).
  * @return {object} Simulator object associated with the udid passed in.
  */
-async function getSimulator (udid) {
+async function getSimulator (udid, opts = {}) {
+  let {
+    platform,
```
mykola-mokhnach: should platform be equal to `iOS` by default?

imurchie: Sure.

PR #257, opened 2020-02-07
refactor: use util.pluralize

imurchie:


imurchie:
Yes, for normal situations, where the last word is pluralized. It fails for the odd situations (e.g., `pluralize('notary public', 4)` incorrectly returns `"notary publics"` instead of `"notaries public"`).

PR #261, opened 2020-02-09
feat: Add a method to list running processes

mykola-mokhnach:
Got an idea from https://apple.stackexchange.com/questions/352107/getting-list-of-running-processes-on-ios-simulator

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -1190,6 +1190,57 @@ class SimulatorXcode6 extends EventEmitter {
   //endregion
 
   /**
+   * @typedef {Object} ProcessInfo
+   * @property {number} pid The actual process identifier.
+   * Could be zero if the process is the system one.
+   * @property {?string} group The process group identifier.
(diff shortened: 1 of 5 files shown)


lib/simulator-xcode-6.js
```diff
@@ -1189,6 +1189,60 @@ class SimulatorXcode6 extends EventEmitter {
...
+   * @throws {Error} if no process information could be retrieved.
+   */
+  async ps () {
+    const {stdout} = await this.simctl.spawnProcess([
+      'launchctl',
+      'print',
+      'system',
+    ]);
+
+    const servicesMatch = /^\s*services\s*=\s*{([^}]+)/m.exec(stdout);
```
KazuCocoa: ` xcrun simctl spawn booted launchctl list`
> list            Lists information about services.

I compared the result and `services = {...}` in the `print system`, then their count was the same. (Not checked one by one)

mykola-mokhnach: could be. I would say both options are fine

PR #262, opened 2020-02-12
feat: disable pasteboard sync in launching time to improve the performance

KazuCocoa:
Reference: 
- https://github.com/WebKit/webkit/blob/master/Tools/Scripts/webkitpy/xcode/simulated_device.py#L413
- https://twitter.com/steipete/status/1227551552317140992?s=20

This argument improves launching simulator performance.
Once this pasteboard sync disabled, the pasteboard cannot sync with macOS system with simulators. In a case a user would like to sync such pasteboard, I'd like to add `disablePasteboardAutomaticSync` as caps.

https://github.com/appium/appium-ios-driver/blob/master/lib/desired-caps.js

Then, `PasteboardAutomaticSync` follows the global preference.

```
defaults read com.apple.iphonesimulator PasteboardAutomaticSync
```

On my machine, this disableness improved the launching performance around two seconds.

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -480,7 +480,7 @@ class SimulatorXcode6 extends EventEmitter {
         // Do not add -PasteboardAutomaticSync
         break;
       default:
-        log.warn(`pasteboardAutomaticSync capability accepts one of ['on', 'off' or 'system']. Follows 'off' as default.`);
+        log.warn(`['on', 'off' or 'system'] are available as the pasteboard automatic sync option. Defaulting to 'off'.`);
         args.push('-PasteboardAutomaticSync', '0');
     }
 
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -457,6 +457,10 @@ class SimulatorXcode6 extends EventEmitter {
       args.push('-ConnectHardwareKeyboard', opts.connectHardwareKeyboard ? '1' : '0');
     }
 
+    // Improve launching simulator performance
+    // https://github.com/WebKit/webkit/blob/master/Tools/Scripts/webkitpy/xcode/simulated_device.py#L413
+    args.push('-PasteboardAutomaticSync', '0');
+
     log.info(`Starting Simulator UI with command: open ${args.join(' ')}`);
     try {
       await exec('open', args, {timeout: opts.startupTimeout});
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -424,21 +424,26 @@ class SimulatorXcode6 extends EventEmitter {
     return indicator;
   }
 
+
+  /**
+   * @typedef {Object} SimulatorOptions
+   * @property {?string} - Defines the window scale value for the UI client window for the current Simulator.
+   *   Equals to null by default, which keeps the current scale unchanged.
+   *   It should be one of ['1.0', '0.75', '0.5', '0.33', '0.25'].
+   * @property {boolean} connectHardwareKeyboard - Whether to connect the hardware keyboard to the
+   *   Simulator UI client. Defaults to false.
+   * @property {boolean} disablePasteboardAutomaticSync - Whether to disable pasteboard sync with the
+   *   Simulator UI client or respect the system wide preference.
+   *   The sync increase launching simulator process time, but it allows system to sync pasteboard
+   *   with simulators. Follows system-wide preference if the value is false.
+   *   Defaults to true.
+   * @property {number} startupTimeout - Number of milliseconds to wait until Simulator booting
+   *   process is completed. The default timeout will be used if not set explicitly.
+   */
+
   /**
    * Start the Simulator UI client with the given arguments
-   *
-   * @param {object} opts - One or more of available Simulator UI client options:
-   *   - {string} scaleFactor: can be one of ['1.0', '0.75', '0.5', '0.33', '0.25'].
-   *   Defines the window scale value for the UI client window for the current Simulator.
-   *   Equals to null by default, which keeps the current scale unchanged.
-   *   - {boolean} connectHardwareKeyboard: whether to connect the hardware keyboard to the
-   *   Simulator UI client. Equals to false by default.
-   *   - {boolean} disablePasteboardAutomaticSync: whether to disable pasteboard sync with the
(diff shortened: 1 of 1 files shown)


lib/simulator-xcode-6.js
```diff
@@ -433,13 +433,18 @@ class SimulatorXcode6 extends EventEmitter {
    *   Equals to null by default, which keeps the current scale unchanged.
```
mykola-mokhnach: would it be possible to move opts type description into a separate `typedef` block, like we do with all complex argument types?

lib/simulator-xcode-6.js
```diff
@@ -457,6 +462,12 @@ class SimulatorXcode6 extends EventEmitter {
       args.push('-ConnectHardwareKeyboard', opts.connectHardwareKeyboard ? '1' : '0');
     }
 
+    if (_.isBoolean(opts.disablePasteboardAutomaticSync)) {
```
mykola-mokhnach: this will not work if opts.disablePasteboardAutomaticSync is set to `false`: the `-PasteboardAutomaticSync` is always set to `0`

lib/simulator-xcode-6.js
```diff
@@ -457,6 +462,12 @@ class SimulatorXcode6 extends EventEmitter {
...
+      // Improve launching simulator performance
+      // https://github.com/WebKit/webkit/blob/master/Tools/Scripts/webkitpy/xcode/simulated_device.py#L413
+      args.push('-PasteboardAutomaticSync', '0');
```
mykola-mokhnach: what is the minimum SDK version where this arg is supported?

lib/simulator-xcode-6.js
KazuCocoa: Oops... true. it's my mistake...

lib/simulator-xcode-6.js
KazuCocoa: It looks Xcode 9.3 simulator has it, at least
image

lib/simulator-xcode-6.js
```diff
@@ -424,22 +424,27 @@ class SimulatorXcode6 extends EventEmitter {
...
+   * @typedef {Object} SimulatorOptions
+   * @property {?string} - Defines the window scale value for the UI client window for the current Simulator.
    *   Equals to null by default, which keeps the current scale unchanged.
-   *   - {boolean} connectHardwareKeyboard: whether to connect the hardware keyboard to the
-   *   Simulator UI client. Equals to false by default.
-   *   - {boolean} disablePasteboardAutomaticSync: whether to disable pasteboard sync with the
-   *   Simulator UI client. It increases launching simulator process time, but it allows system
-   *   to sync pasteboard with simulators. Follows system-wide preference if the value is false.
+   *   It should be one of ['1.0', '0.75', '0.5', '0.33', '0.25'].
+   * @property {boolean} connectHardwareKeyboard - Whether to connect the hardware keyboard to the
```
mykola-mokhnach: simply put `connectHardwareKeyboard [false]`

lib/simulator-xcode-6.js
```diff
@@ -462,8 +467,9 @@ class SimulatorXcode6 extends EventEmitter {
       args.push('-ConnectHardwareKeyboard', opts.connectHardwareKeyboard ? '1' : '0');
     }
 
-    if (_.isBoolean(opts.disablePasteboardAutomaticSync)) {
-      // Improve launching simulator performance
+    if (_.isBoolean(opts.disablePasteboardAutomaticSync) && opts.disablePasteboardAutomaticSync) {
+      // Improve launching simulator performance.
+      // Respect system wide preference if disablePasteboardAutomaticSync was false.
       // https://github.com/WebKit/webkit/blob/master/Tools/Scripts/webkitpy/xcode/simulated_device.py#L413
       args.push('-PasteboardAutomaticSync', '0');
```
mykola-mokhnach: how about

```
args.push('-PasteboardAutomaticSync', `${+opts.disablePasteboardAutomaticSync}`)
```

lib/simulator-xcode-6.js
mykola-mokhnach: I was just thinking about this option is available since simulator9 and we have the code inside simulator6, but I would say additional args does not affect older simulators functionality anyway.

lib/simulator-xcode-6.js
```diff
@@ -424,22 +424,27 @@ class SimulatorXcode6 extends EventEmitter {
...
+   *   with simulators. Follows system-wide preference if the value is false.
    *   Defaults to true.
-   *   - {number} startupTimeout: number of milliseconds to wait until Simulator booting
+   * @property {number} startupTimeout - Number of milliseconds to wait until Simulator booting
    *   process is completed. The default timeout will be used if not set explicitly.
    */
+
+  /**
+   * Start the Simulator UI client with the given arguments
+   * @param {Object} SimulatorOptions - Simulator launching options
```
mykola-mokhnach: `@param {SimulatorOptions} opts`

lib/simulator-xcode-6.js
```diff
@@ -424,22 +424,27 @@ class SimulatorXcode6 extends EventEmitter {
...
 
+
   /**
-   * Start the Simulator UI client with the given arguments
-   *
-   * @param {object} opts - One or more of available Simulator UI client options:
-   *   - {string} scaleFactor: can be one of ['1.0', '0.75', '0.5', '0.33', '0.25'].
-   *   Defines the window scale value for the UI client window for the current Simulator.
+   * @typedef {Object} SimulatorOptions
+   * @property {?string} - Defines the window scale value for the UI client window for the current Simulator.
```
mykola-mokhnach: `@property {?string} scaleFactor`

lib/simulator-xcode-6.js
KazuCocoa: I considered the way, but I kept a case not to apply `PasteboardAutomaticSync` as launching argument since it could set in system-wide preference by `defaults write com.apple.iphonesimulator PasteboardAutomaticSync -bool false`.
Or `pasteboardAutomaticSync: 'on'`, `pasteboardAutomaticSync: 'off'`, `pasteboardAutomaticSync: 'system'`?
(Just came up with this idea)

'on': force it on
'off': force it off (current approach)
'system': does not set this flag

mykola-mokhnach: true/false/(null|undefined) would serve the same approach

PR #273, opened 2020-03-31
chore: fix linting issues

imurchie:


lib/tail-until.js
```diff
@@ -12,15 +12,13 @@ async function tailUntil (filePath, until, timeout = 5000) {
   // }
   // proc.on('output', consoleOut);
 
-  let startDetector = (stdout) => {
-    return stdout.indexOf(until) > -1;
-  };
+  const startDetector = (stdout) => stdout.indexOf(until) > -1;
```
mykola-mokhnach: could use includes

PR #281, opened 2020-06-12
refactor: Prefer own implementation of app activation over idb one

mykola-mokhnach:
I've checked how `focus` command is implemented in idb code and ported it to PyObjC

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -306,7 +306,7 @@ async function getDeveloperRoot () {
  * See https://developer.apple.com/documentation/appkit/nsrunningapplication/1528725-activatewithoptions?language=objc
  * for more details.
  *
- * @param {number|string} pid App identifier
+ * @param {number|string} pid App process identifier
  * @throws {Error} If the given PID is not running or there was a failure
  * while activating the app
  */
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -289,6 +301,28 @@ async function getDeveloperRoot () {
...
+ * while activating the app
+ */
+async function activateApp (pid) {
+  const tmpScript = await tempDir.path({
+    prefix: `activate_sim_${util.uuidV4().substring(0, 8)}`,
+    suffix: '.py',
+  });
+  await fs.writeFile(tmpScript, APP_ACTIVATION_SCRIPT(pid), 'utf8');
+  try {
+    await exec('/usr/bin/python', [tmpScript]);
```
KazuCocoa, line 320: how about catch and print an error to install https://pypi.org/project/pyobjc/ ?
No module error will be raised if the machine has no PyObjC

mykola-mokhnach: Python with ObjC bindings is always included into MacOS by default

KazuCocoa: Okay
/usr/local/lib/python2.7/site-packages/ had AppKit, but system installed python3.x did not have. So I wondered I manually installed it in my past.

mykola-mokhnach: https://stackoverflow.com/questions/6819661/python-location-on-mac-osx

PR #282, opened 2020-06-15
fix: sim window minized on xcode 11.4

sadafyadegari:
Simulators remain minimized when running tests with Xcode 11.4.1. Fix was to launch simulator UI client before simulators are booted.

### DIFF
--- a/lib/simulator-xcode-11.4.js
+++ b/lib/simulator-xcode-11.4.js
@@ -114,9 +114,8 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
     if (!isUiClientRunning) {
       await this.startUIClient(opts);
     }
-    await this.bootSimulator();
+    await this.boot();
   }
-
 
 }
 
--- a/lib/simulator-xcode-11.4.js
+++ b/lib/simulator-xcode-11.4.js
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import SimulatorXcode11 from './simulator-xcode-11';
 import log from './logger';
+import {retryInterval} from "asyncbox";
 
 
 class SimulatorXcode11_4 extends SimulatorXcode11 {
@@ -107,22 +108,18 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
   }
 
   /**
-   * Opens simulator's UI Client if not already open and boots sim
-   *
-   * @param uiClientPid - process id of simulator UI client.
-   * @param opts - arguments to start simulator UI client with.
-   * @returns {Promise} - a promise that resolves when sim UI client is launched and sim is booted.
+   * @inheritdoc
    * @override
-   */
-  async launchSimWindow (uiClientPid, opts) {
+   * */
+  async launchWindow (isUiClientRunning, opts) {
     // In xcode 11.4, UI Client must be first launched, otherwise
     // sim window stays minimized
-    if (!uiClientPid) {
+    if (!isUiClientRunning) {
       await this.startUIClient(opts);
     }
-    log.info(`Booting Simulator with UDID '${this.udid}'...`);
     await this.bootSimulator();
   }
+
 
 }
 
--- a/lib/simulator-xcode-11.4.js
+++ b/lib/simulator-xcode-11.4.js
@@ -1,8 +1,5 @@
 import _ from 'lodash';
 import SimulatorXcode11 from './simulator-xcode-11';
-import log from './logger';
-import {retryInterval} from "asyncbox";
-
 
 class SimulatorXcode11_4 extends SimulatorXcode11 {
   constructor (udid, xcodeVersion) {
--- a/lib/simulator-xcode-11.4.js
+++ b/lib/simulator-xcode-11.4.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import SimulatorXcode11 from './simulator-xcode-11';
+import log from './logger';
 
 
 class SimulatorXcode11_4 extends SimulatorXcode11 {
@@ -104,6 +105,25 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
   async clearKeychains () {
     await this.simctl.resetKeychain();
   }
+
+  /**
+   * Opens simulator's UI Client if not already open and boots sim
+   *
+   * @param uiClientPid - process id of simulator UI client.
+   * @param opts - arguments to start simulator UI client with.
+   * @returns {Promise} - a promise that resolves when sim UI client is launched and sim is booted.
+   * @override
+   */
+  async launchSimWindow (uiClientPid, opts) {
+    // In xcode 11.4, UI Client must be first launched, otherwise
+    // sim window stays minimized
+    if (!uiClientPid) {
+      await this.startUIClient(opts);
+    }
+    log.info(`Booting Simulator with UDID '${this.udid}'...`);
+    await this.bootSimulator();
+  }
+
 }
 
 export default SimulatorXcode11_4;
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -138,7 +138,7 @@ class SimulatorXcode9 extends SimulatorXcode8 {
           }
           await waitForShutdown();
         }
-        await this.launchSimWindow(uiClientPid, opts);
+        await this.launchWindow(uiClientPid, opts);
       }
       return true;
     });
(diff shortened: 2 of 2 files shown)


KazuCocoa:
Thank you!
Could you sign the CLA?

sadafyadegari:
Just signed it 👍

lib/simulator-xcode-11.4.js
```diff
@@ -104,6 +105,25 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
   async clearKeychains () {
     await this.simctl.resetKeychain();
   }
+
+  /**
+   * Opens simulator's UI Client if not already open and boots sim
+   *
+   * @param uiClientPid - process id of simulator UI client.
```
mykola-mokhnach, line 112: You could simply mention `@inherit` if the documentation repeats the overridden one

lib/simulator-xcode-11.4.js
```diff
@@ -104,6 +105,25 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
...
+   * @param opts - arguments to start simulator UI client with.
+   * @returns {Promise} - a promise that resolves when sim UI client is launched and sim is booted.
+   * @override
+   */
+  async launchSimWindow (uiClientPid, opts) {
```
mykola-mokhnach, line 117: I would rather change the `uiClientPid` argument to boolean `isUiClientRunning`

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,34 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
   }
 
+  /***
+   * Boots simulator and opens simulators UI Client if not already opened.
+   *
+   * @param uiClientPid - process id of simulator UI client.
+   * @param opts - arguments to start simulator UI client with.
+   * @returns {Promise} - a promise that resolves when sim is booted and sim UI client is launched.
+   */
+  async launchSimWindow (uiClientPid, opts) {
```
mykola-mokhnach, line 159: these method don't need the word `Sim` or `Simulator` in them, since the class already has it

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,34 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     }
   }
 
+  /***
+   * Boots simulator and opens simulators UI Client if not already opened.
+   *
+   * @param uiClientPid - process id of simulator UI client.
```
mykola-mokhnach, line 155: Please mention the argument types

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,34 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+   * @param opts - arguments to start simulator UI client with.
+   * @returns {Promise} - a promise that resolves when sim is booted and sim UI client is launched.
```
mykola-mokhnach, line 157: the return description is not needed - the method does not return nothing

lib/simulator-xcode-11.4.js
```diff
@@ -104,6 +105,25 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
...
+    // In xcode 11.4, UI Client must be first launched, otherwise
+    // sim window stays minimized
+    if (!uiClientPid) {
+      await this.startUIClient(opts);
+    }
+    log.info(`Booting Simulator with UDID '${this.udid}'...`);
```
mykola-mokhnach, line 123: this log line could be moved to `boot` method

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,34 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+  }
+
+  /**
+   * Boots simulator.
+   *
+   * @returns {Promise} - A promise that resolved when simulator is booted.
+   */
+  async bootSimulator () {
+    try {
+      await retryInterval(3, 2000, async () => await this.simctl.bootDevice());
```
mykola-mokhnach, line 174: this call might cause unnecessary delays if the simulator is already booted. The original command being executed more than once returns non-zero exit code:

```
An error was encountered processing the command (domain=com.apple.CoreSimulator.SimError, code=165):
Unable to boot device in current state: Booted
```

lib/simulator-xcode-11.4.js
sadafyadegari: sounds good, using `@inheritdoc tag` that also implies override. https://jsdoc.app/tags-inheritdoc.html

lib/simulator-xcode-11.4.js
sadafyadegari: Makes sense, I changed it.

lib/simulator-xcode-9.js
sadafyadegari: Ok, will remove the `Sim`s and `Simulators`s .

lib/simulator-xcode-9.js
sadafyadegari: Will do!

lib/simulator-xcode-9.js
sadafyadegari: 👍

lib/simulator-xcode-11.4.js
sadafyadegari: Good catch, moved it.

lib/simulator-xcode-9.js
sadafyadegari: Ok that's a good point, I scanned [run(opts = {})](https://github.com/sadafyadegari/appium-ios-simulator/blob/master/lib/simulator-xcode-9.js#L67) and it shouldn't call BootSimulator when `serverState === 'Booted'` . But I agree that it'd be good to add a guard against that case for posterity. I'll add a sim state check inside `bootSimulator` function.

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,37 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+    if (!isUiClientRunning) {
+      await this.startUIClient(opts);
+    }
+  }
+
+  /**
+   * Boots simulator if not already booted.
+   */
+  async bootSimulator () {
+    const {state: serverState} = await this.stat();
```
mykola-mokhnach, line 169: I would rather catch the exception from `simctl.bootDevice` and check stderr to decide whether to retry or not. Calling `stat` adds a performance hit.

sadafyadegari: @mykola-mokhnach are you looking for something like 
`async bootSimulator () {
    try {
      log.info(`Booting Simulator with UDID '${this.udid}'...`);
      await this.simctl.bootDevice();
    } catch (err) {
      if (!(err.stdout || '').includes('Unable to boot device in current state: Booted') &&
          !(err.stderr || '').includes('Unable to boot device in current state: Booted')) {
        await retryInterval(3, 2000, async () => await this.simctl.bootDevice());
      } else {
        log.warn(`Error while booting simulator: ${err.stdout || err.stderr}.`);
      }
    }
  }`
?

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+    await this.bootSimulator();
+    if (!isUiClientRunning) {
+      await this.startUIClient(opts);
+    }
+  }
+
+  /**
+   * Boots simulator if not already booted.
+   */
+  async bootSimulator () {
```
mykola-mokhnach, line 168: Please rename to simply `boot`

lib/simulator-xcode-11.4.js
```diff
@@ -104,6 +103,21 @@ class SimulatorXcode11_4 extends SimulatorXcode11 {
...
+  async launchWindow (isUiClientRunning, opts) {
+    // In xcode 11.4, UI Client must be first launched, otherwise
+    // sim window stays minimized
+    if (!isUiClientRunning) {
+      await this.startUIClient(opts);
+    }
+    await this.bootSimulator();
+  }
+
+
```
mykola-mokhnach, line 120: extra line break

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     }
   }
 
+  /***
+   * Boots simulator and opens simulators UI Client if not already opened.
+   *
+   * @param {?String} isUiClientRunning - process id of simulator UI client.
```
mykola-mokhnach, line 155: `{boolean}`

lib/simulator-xcode-9.js
```diff
@@ -145,11 +138,7 @@ class SimulatorXcode9 extends SimulatorXcode8 {
           }
           await waitForShutdown();
         }
-        log.info(`Booting Simulator with UDID '${this.udid}'...`);
-        await bootSimulator();
-        if (!uiClientPid) {
-          await this.startUIClient(opts);
-        }
+        await this.launchWindow(uiClientPid, opts);
```
mykola-mokhnach, line 141: !!uiClientPid

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     }
   }
 
+  /***
+   * Boots simulator and opens simulators UI Client if not already opened.
+   *
+   * @param {?String} isUiClientRunning - process id of simulator UI client.
+   * @param {Object} opts - arguments to start simulator UI client with.
```
mykola-mokhnach, line 156: {RunOptions}

lib/simulator-xcode-9.js
```diff
@@ -160,6 +149,40 @@ class SimulatorXcode9 extends SimulatorXcode8 {
...
+   */
+  async launchWindow (isUiClientRunning, opts) {
```
mykola-mokhnach, line 158: opts = {}

lib/simulator-xcode-9.js
KazuCocoa: Is this the last item?

PR #284, opened 2020-08-20
refactor: Change simulator state detection logic to use faster APIs

mykola-mokhnach:
The `stat` method internally invokes `simctl list`, which could be pretty slow, especially if multiple simulators are installed on the system. I've switched the state detection API to use `getenv` call as a workaround. In opposite to the `list` call it only queries single simulator, which makes the overall performance better. On my machine the timings were:

```
$ time xcrun simctl getenv 0B2C9CE6-29AC-4494-B64F-CE711619146E dummy
...
xcrun simctl getenv 0B2C9CE6-29AC-4494-B64F-CE711619146E dummy  0.08s user 0.09s system 88% cpu 0.184 total
```

```
$ % time xcrun simctl list
....
xcrun simctl list  0.11s user 0.20s system 43% cpu 0.723 total
```

### DIFF
--- a/lib/simulator-xcode-6.js
+++ b/lib/simulator-xcode-6.js
@@ -364,8 +364,29 @@ class SimulatorXcode6 extends EventEmitter {
    * @return {boolean} True if the current Simulator is running.
    */
   async isRunning () {
-    let stat = await this.stat();
-    return stat.state === 'Booted';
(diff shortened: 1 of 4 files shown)


mykola-mokhnach:
cc @sauce-tartar

PR #288, opened 2020-09-16
fix: use simctl privacy for calendar access

dpgraham:
Makes a call to `xcrun simctl privacy...` instead of directly editing TCC.db to grant calendar access.

Also changes the simulator.js logic so 
1. it explicitly throws an error if Xcode versions 1 ~> 5 are provided.
2. if anything >= 11.4 is provided, it will default to `simulator-xcode-11.4` implementation

_WHY?_ So that when Xcode 13 comes out it will, by default, use the latest Xcode implementation. And if it works (ie: our e2e tests keep passing) we don't need to do anything

(depends on https://github.com/appium/node-simctl/pull/114 ... make sure to publish this first and bump up min node-simctl before merging)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,9 @@ language: node_js
 node_js: "10"
 matrix:
   include:
-    - osx_image: xcode7.3
-      env: DEVICE=9.3
-
-    - osx_image: xcode8.3
-      env: DEVICE=9.3
-    - osx_image: xcode8.3
-      env: DEVICE=10.3
-
-    - osx_image: xcode9.3
-      env: DEVICE=10.3
     - osx_image: xcode9.3
       env: DEVICE=11.3
 
-    # TODO: uncomment once Travis 9.4 VMs have iOS 11.3 SDK installed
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -3,22 +3,9 @@ language: node_js
...
-
-    - osx_image: xcode9.3
-      env: DEVICE=10.3
     - osx_image: xcode9.3
       env: DEVICE=11.3
 
-    # TODO: uncomment once Travis 9.4 VMs have iOS 11.3 SDK installed
-    # - osx_image: xcode9.4
-    #   env: DEVICE=11.3
     - osx_image: xcode9.4
```
mykola-mokhnach, line 9: how about adding new environments instead?

mykola-mokhnach (approved):

.travis.yml
dpgraham: @mykola-mokhnach yeah when I get a chance I'll put in a new PR and port this stuff over to Azure.

PR #305, opened 2021-04-05
fix: idb and AppleScript Geo Location setters

nextlevelbeard:
This fixes two issues with `setGeoLocation`:
1. The `idb` location setter is throwing an error even on success, making the fallback `AppleScript` location setter always run
1. The `AppleScript` location setter fails to set valid coordinates for certain locales due to a wrong calculation of the decimal separator

   To test:
   1. On macOS go to `Language and Region` → `System Preferences` and either change your locale settings or change the separator by going to `Advanced` → `Number separators` → `Decimal`
   
      walkthrough 1 walkthrough 2
   1. Run the script to see your changes reflected:
      ```shell
      osascript -e 'use framework "Foundation"' -e 'return (decimalSeparator of init() of alloc() of NSNumberFormatter of current application as string)'
      ```
   1. Open Simulator,  enter Custom Location /w the new separator: `Feature` → `Location` → `Custom Location`

### DIFF
--- a/lib/geolocation.js
+++ b/lib/geolocation.js
@@ -53,8 +53,7 @@ async function setLocationWithIdb (idb, latitude, longitude) {
     } catch (e) {
       throw new Error(`Failed to set geolocation with idb. Original error: ${e.stderr || e.message}`);
     }
-  }
-  else {
+  } else {
     throw new Error('Failed to set geolocation with idb because it is not installed or the "launchWithIDB" capability was not set');
   }
 }
@@ -73,12 +72,14 @@ async function setLocationWithIdb (idb, latitude, longitude) {
  */
 async function setLocationWithAppleScript (sim, latitude, longitude, menu = 'Debug') {
   // Make sure system-wide decimal separator is used
-  const decimalSeparator = (await sim.executeUIClientScript(`
-    use framework "Foundation"
-    return (decimalSeparator of init() of alloc() of NSNumberFormatter of current application as string)
-  `)).trim();
+  const decimalSeparator = (await exec('/usr/bin/python', [
+    '-c',
+    'from AppKit import NSNumberFormatter;' +
+    'import sys;' +
+    'sys.stdout.write(NSNumberFormatter.alloc().init().decimalSeparator())'
+  ])).stdout;
 
-  const [latitudeStr, longitudeStr] = [latitude, longitude].map(s => s.replace(/\D/g, decimalSeparator))
+  const [latitudeStr, longitudeStr] = [latitude, longitude].map((coord) => `${coord}`.replace(/\D/g, decimalSeparator));
 
   const output = await sim.executeUIClientScript(`
     tell application "System Events"
--- a/lib/geolocation.js
+++ b/lib/geolocation.js
@@ -54,7 +54,9 @@ async function setLocationWithIdb (idb, latitude, longitude) {
       throw new Error(`Failed to set geolocation with idb. Original error: ${e.stderr || e.message}`);
     }
   }
-  throw new Error('Failed to set geolocation with idb because it is not installed');
+  else {
+    throw new Error('Failed to set geolocation with idb because it is not installed');
+  }
 }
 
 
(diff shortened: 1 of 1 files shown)


lib/geolocation.js
```diff
@@ -70,10 +72,14 @@ async function setLocationWithIdb (idb, latitude, longitude) {
  * @throws {Error} If it failed to set the location
  */
 async function setLocationWithAppleScript (sim, latitude, longitude, menu = 'Debug') {
-  const decimalSeparator = (0.1).toLocaleString().replace(/\d/g, '');
   // Make sure system-wide decimal separator is used
-  const latitudeStr = `${latitude}`.replace(/\D/g, decimalSeparator);
-  const longitudeStr = `${longitude}`.replace(/\D/g, decimalSeparator);
+  const decimalSeparator = (await sim.executeUIClientScript(`
```
mykola-mokhnach, line 76: this might cause security alert. Try to use `exec` with `/usr/bin/python -c "from AppKit import NSNumberFormatter;import sys; sys.stdout.write(NSNumberFormatter.alloc().init().decimalSeparator())"` command

lib/geolocation.js
```diff
@@ -54,7 +54,9 @@ async function setLocationWithIdb (idb, latitude, longitude) {
       throw new Error(`Failed to set geolocation with idb. Original error: ${e.stderr || e.message}`);
     }
   }
-  throw new Error('Failed to set geolocation with idb because it is not installed');
+  else {
```
mykola-mokhnach, line 57: please move `else` to the same line where the closing bracket is located

lib/geolocation.js
```diff
@@ -70,10 +72,14 @@ async function setLocationWithIdb (idb, latitude, longitude) {
...
-  const decimalSeparator = (0.1).toLocaleString().replace(/\d/g, '');
   // Make sure system-wide decimal separator is used
-  const latitudeStr = `${latitude}`.replace(/\D/g, decimalSeparator);
-  const longitudeStr = `${longitude}`.replace(/\D/g, decimalSeparator);
+  const decimalSeparator = (await sim.executeUIClientScript(`
+    use framework "Foundation"
+    return (decimalSeparator of init() of alloc() of NSNumberFormatter of current application as string)
+  `)).trim();
+
+  const [latitudeStr, longitudeStr] = [latitude, longitude].map(s => s.replace(/\D/g, decimalSeparator))
```
mykola-mokhnach, line 81: Please use the following syntax for lambdas: `(s) => s.replace(/\D/g, decimalSeparator)`

mykola-mokhnach: s.replace will fail if latitude or longitude is a number

lib/geolocation.js
nextlevelbeard: Can we test whether or not it does? Had no issues here on latest macOS.
Feel like we should be very careful here about introducing a `python` dependency.

mykola-mokhnach: python is a part of macOS (at least the one under `/usr/bin`). This command does exactly the same that osascript does, just without any overhead

nextlevelbeard: good to know, thanks!

lib/geolocation.js
nextlevelbeard: good catch, linted and fixed

lib/geolocation.js
```diff
@@ -53,8 +53,9 @@ async function setLocationWithIdb (idb, latitude, longitude) {
     } catch (e) {
       throw new Error(`Failed to set geolocation with idb. Original error: ${e.stderr || e.message}`);
     }
+  } else {
```
mykola-mokhnach, line 56: I think it would be more readable to turn this into a guard and move to the beginning of the method:

```
if (!idb) {
  throw ...
}
```

mykola-mokhnach (approved):

lib/geolocation.js
nextlevelbeard: Agreed

nextlevelbeard:
@mykola-mokhnach Think you can investigate the CI failures, possibly rerun failed jobs?

mykola-mokhnach:
this looks like some temporary glitch. the PR is ok, thanks

PR #298, opened 2020-12-09
fix: Update simulator preferences handling

mykola-mokhnach:
Plist preference files on macOS are managed by cfprefsd. Changing the file itself without letting the daemon know about it has no effect. Read https://eclecticlight.co/2017/07/06/sticky-preferences-why-trashing-or-editing-them-may-not-change-anything/ and https://shadowfile.inode.link/blog/2018/06/advanced-defaults1-usage/ for more details

PR #307, opened 2021-05-06
feat: Add a possibility to change locale settings without restarting the Simulator

mykola-mokhnach:
The idea is similar to https://github.com/appium/appium-ios-simulator/pull/298 except of Simulator has its own `defaults` tool, which could be triggered via `simctl spawn`

### DIFF
--- a/test/functional/simulator-e2e-specs.js
+++ b/test/functional/simulator-e2e-specs.js
@@ -370,6 +370,28 @@ describe('advanced features', function () {
     });
   });
 
+  describe('configureLocalization', function () {
+    it(`should properly set locale settings`, async function () {
+      if (!_.isFunction(sim.configureLocalization)) {
+        return this.skip();
+      }
+
+      await sim.configureLocalization({
+        language: {
+          name: 'en'
+        },
+        locale: {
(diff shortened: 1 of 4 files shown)


test/unit/defaults-utils-specs.js
```diff
@@ -64,6 +64,18 @@ describe('defaults-utils', function () {
       ]);
     });
 
+    it('could properly generate command args for value types with replacement', function () {
+      generateDefaultsCommandArgs({
```
KazuCocoa, line 68: wow, great to find

test/functional/simulator-e2e-specs.js
```diff
@@ -370,6 +370,28 @@ describe('advanced features', function () {
...
+        return this.skip();
+      }
+
+      await sim.configureLocalization({
+        language: {
+          name: 'en'
+        },
+        locale: {
+          name: 'en_US',
+          calendar: 'gregorian',
```
KazuCocoa, line 385: Awesome...
Btw, non-gregorian like japanese one is a typical calendar type which cause issues.

KazuCocoa (approved):

PR #322, opened 2022-04-06
fix: Start Safari instead of loading appium.io

mykola-mokhnach:


jlipps (approved):

lib/simulator-xcode-6.js
```diff
@@ -643,7 +643,7 @@ class SimulatorXcode6 extends EventEmitter {
     await this.run({startupTimeout});
 
     if (safari) {
-      await this.openUrl('http://www.appium.io');
```
KazuCocoa, line 646: maybe this does not work since `openurl` command expected http/https url... so we launched exact url in some places

jlipps: ah yeah i think we ran into that before. that is one reason we hosted had the page try to load the appium server's own demo page itself at one point.

mykola-mokhnach: > maybe this does not work since `openurl` command expected http/https url... so we launched exact url in some places

Thanks for the head up @KazuCocoa 
It turns out `simctl openurl` really does not like urls without protocol. I've changed the logic to simply open the browser and verify it's running instead

KazuCocoa (approved):
looks good, like it

PR #342, opened 2022-12-09
feat: add setReduceTransparency

rerorero:
This PR adds setReduceTransparency function that allows to switch `Reduce Transparency` in accessibility preference.
Please see  https://github.com/appium/appium/issues/17905 for the background, and feel free to close this if this feature request doesn't make sense.

I found `ReduceMotion` is implemented with a different approach https://github.com/appium/WebDriverAgent/pull/202
I'm not sure which way is better though, I confirmed `EnhancedBackgroundContrastEnabled` can be changed in runtime and reflected immediately. So I chose this way but please let me know if WDA is better then I'll try to do it.

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -157,6 +157,16 @@ async function setReduceMotion (sim, reduceMotion = true) {
   }
 }
 
+async function setReduceTransparency (sim, reduceTransparency = true) {
+  log.debug(`Updating reduce tranceparency. Setting to ${reduceTransparency}.`);
+  const paths = await plistPaths(sim, 'accessibilitySettings');
+  for (const file of paths) {
+    await update(file, {
+      EnhancedBackgroundContrastEnabled: reduceTransparency ? 1 : 0
+    });
+  }
+}
+
 async function updateSafariGlobalSettings (sim, settingSet) {
   log.debug('Updating Safari global settings');
 
@@ -270,7 +280,7 @@ async function stub () {
 }
 
 export {
-  update, updateSettings, updateLocationSettings, setReduceMotion,
+  update, updateSettings, updateLocationSettings, setReduceMotion, setReduceTransparency,
   updateSafariUserSettings, updateSafariGlobalSettings, updateLocale, read,
   readSettings, stub,
 };
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -157,7 +157,7 @@ async function setReduceMotion (sim, reduceMotion = true) {
   }
 }
 
-async function setReduceTransparency (sim, reduceTransparency = true) {
+async function setReduceTransparency (sim, reduceTransparency) {
   log.debug(`Updating reduce tranceparency. Setting to ${reduceTransparency}.`);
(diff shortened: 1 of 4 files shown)


lib/simulator-xcode-6.js
```diff
@@ -772,6 +772,19 @@ class SimulatorXcode6 extends EventEmitter {
...
   }
 
+  /**
+   * Enable/Disable reduce transparency.
+   *
+   * @param {boolean} reduceMotion - Whether or not to enable it.
+   */
+  async setReduceTransparency (reduceTransparency = true) {
+    if (await this.isFresh()) {
+      await this.launchAndQuit(false, STARTUP_TIMEOUT);
```
mykola-mokhnach, line 782: I would say it is an unexpected side effect to have this call in the setter. I would rather remove it (as well as from `setReduceMotion`) and let the caller to decide

lib/settings.js
```diff
@@ -157,6 +157,16 @@ async function setReduceMotion (sim, reduceMotion = true) {
   }
 }
 
+async function setReduceTransparency (sim, reduceTransparency = true) {
+  log.debug(`Updating reduce tranceparency. Setting to ${reduceTransparency}.`);
+  const paths = await plistPaths(sim, 'accessibilitySettings');
```
mykola-mokhnach, line 162: oh, these hardcoded identifiers. It needs to be refactored later.

lib/simulator-xcode-6.js
rerorero: Sure, I'll remove it.

lib/simulator-xcode-6.js
```diff
@@ -772,6 +772,15 @@ class SimulatorXcode6 extends EventEmitter {
     await settings.setReduceMotion(this, reduceMotion);
   }
 
+  /**
+   * Enable/Disable reduce transparency.
+   *
+   * @param {boolean} reduceMotion - Whether or not to enable it.
+   */
+  async setReduceTransparency (reduceTransparency = true) {
```
mykola-mokhnach, line 780: it also makes sense to remove the default value there

lib/settings.js
```diff
@@ -157,6 +157,16 @@ async function setReduceMotion (sim, reduceMotion = true) {
   }
 }
 
+async function setReduceTransparency (sim, reduceTransparency = true) {
```
mykola-mokhnach, line 160: please remove the default value

lib/simulator-xcode-6.js
rerorero: fixed https://github.com/appium/appium-ios-simulator/pull/342/commits/4140047b29e3772e22e2d67032b9e8812b54aec5

lib/settings.js
rerorero: I'll take a look at it once this gets merged.

mykola-mokhnach (approved):

mykola-mokhnach:
I don't think there would be a need to frequently change this setting while a session is running

PR #347, opened 2022-12-12
fix: Update Simulator preferences configuration

mykola-mokhnach:
There is no need to restart Simulator in order to apply its preferences (and it does not work as well).
The right way of doing that now is by invoking `defaults` CLI. It also notifies cfprefsd, so various apps and services could pick up settings change in runtime. 

cc @rerorero


jlipps (approved):
LGTM but I don't see anything about not restarting the sim anymore?

mykola-mokhnach:
> LGTM but I don't see anything about not restarting the sim anymore?

It's done in xcuitest driver. I'll change the implementation as soon as this patch is published

PR #351, opened 2022-12-19
fix: add ios 14 and above case

KazuCocoa:
This PR is to fix getAcccess for iOS 14+ after https://github.com/appium/appium-ios-simulator/pull/350

https://github.com/wix/AppleSimulatorUtils/blob/5467857407da4653305af119b2486051664a4f97/applesimutils/applesimutils/SetServicePermission.m

Should handle iOS version, or probably try iOs 14's first then fallback to existing one if an exception occurs only

### DIFF
--- a/lib/extensions/permissions.js
+++ b/lib/extensions/permissions.js
@@ -110,11 +110,10 @@ async function setAccess (udid, bundleId, permissionsMapping) {
 async function getAccess (bundleId, serviceName, simDataRoot) {
   const internalServiceName = toInternalServiceName(serviceName);
   const dbPath = path.resolve(simDataRoot, 'Library', 'TCC', 'TCC.db');
-
   const getAccessStatus = async (statusPairs, statusKey) => {
     for (const [statusValue, status] of statusPairs) {
       const sql = `SELECT count(*) FROM 'access' ` +
-      `WHERE client='${bundleId}' AND ${statusKey}=${statusValue} AND service='${internalServiceName}'`;
+        `WHERE client='${bundleId}' AND ${statusKey}=${statusValue} AND service='${internalServiceName}'`;
       const count = await execSQLiteQuery(dbPath, sql);
       if (parseInt(count.split('=')[1], 10) > 0) {
         return status;
--- a/lib/extensions/permissions.js
+++ b/lib/extensions/permissions.js
@@ -111,26 +111,31 @@ async function getAccess (bundleId, serviceName, simDataRoot) {
   const internalServiceName = toInternalServiceName(serviceName);
   const dbPath = path.resolve(simDataRoot, 'Library', 'TCC', 'TCC.db');
 
-  // iOS 14 and above
-  for (const [sqlValue, status] of [['0', STATUS_NO], ['2', STATUS_YES], ['3', STATUS_LIMITED]]) {
-    const sql = `SELECT count(*) FROM 'access' ` +
-    `WHERE client='${bundleId}' AND auth_value=${sqlValue} AND service='${internalServiceName}'`;
-    const count = await execSQLiteQuery(dbPath, sql);
-    if (parseInt(count.split('=')[1], 10) > 0) {
-      return status;
+  const getAcccessStatus = async (statusPairs, statusKey) => {
+    for (const [statusValue, status] of statusPairs) {
+      const sql = `SELECT count(*) FROM 'access' ` +
+      `WHERE client='${bundleId}' AND ${statusKey}=${statusValue} AND service='${internalServiceName}'`;
+      const count = await execSQLiteQuery(dbPath, sql);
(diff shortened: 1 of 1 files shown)


lib/extensions/permissions.js
```diff
@@ -78,18 +79,19 @@ async function execWix (args) {
...
  *
+ * @param {string} udid - udid of the target simulator device.
  * @param {string} bundleId - bundle identifier of the target application.
  * @param {Object} permissionsMapping - An object, where keys ar  service names
  * and values are corresponding state values. See https://github.com/wix/AppleSimulatorUtils
  * for more details on available service names and statuses.
  * @throws {Error} If there was an error while changing permissions.
  */
-async function setAccess (bundleId, permissionsMapping) {
+async function setAccess (udid, bundleId, permissionsMapping) {
```
mykola-mokhnach, line 89: please rebase with master

lib/extensions/permissions.js
```diff
@@ -102,21 +104,37 @@ async function setAccess (bundleId, permissionsMapping) {
...
+      if (parseInt(count.split('=')[1], 10) > 0) {
+        return status;
+      }
     }
+    return STATUS_UNSET;
+  };
+
+  try {
+    // iOS 14 and above
+    return await getAccessStatus(
```
mykola-mokhnach, line 127: shall we instead just explicitly check the xcode version?

lib/extensions/permissions.js
```diff
@@ -102,21 +104,37 @@ async function setAccess (bundleId, permissionsMapping) {
  * @param {string} serviceName - the name of the service. Should be one of
  * `SERVICES` keys.
  * @param {string} simDataRoot - the path to Simulator `data` root
- * @returns {string} - The current status: yes/no/unset
+ * @returns {string} - The current status: yes/no/unset/limited
```
mykola-mokhnach, line 107: the documentation in xcuitest-driver should be updated after this change is merged

lib/extensions/permissions.js
KazuCocoa: Yea, potentially. I need to check various environment

lib/extensions/permissions.js
```diff
@@ -7,6 +7,7 @@ import path from 'path';
 const STATUS_UNSET = 'unset';
 const STATUS_NO = 'no';
 const STATUS_YES = 'yes';
+const STATUS_LIMITED = 'limited';
```
mykola-mokhnach, line 10: I assume it would be more readable to create a mapping like 
```
const STATUS = Object.freeze({
  YES: 'yes',
  ....
})
```

lib/extensions/permissions.js
KazuCocoa: Perhaps... Xcode 13 can work. Xcode 12 env did not on my local

KazuCocoa: hm, it seems the DB is not Xcode version dependent. iOS 14 created in Xcode 12 had `auth_value`, but did not iOS 12

lib/extensions/permissions.js
KazuCocoa: yes

lib/extensions/permissions.js
```diff
@@ -103,24 +107,40 @@ async function setAccess (udid, bundleId, permissionsMapping) {
...
     }
+    return STATUS.UNSET;
+  };
+
+  // 'auth_value' existence depends on the OS version rather than Xcode version.
+  // Thus here check the newer one first, then fallback to the older version way.
+  try {
+    // iOS 14+
+    return await getAccessStatus(
+      [['0', STATUS.NO], ['2', STATUS.YES], ['3', STATUS.LIMITED]],
```
mykola-mokhnach, line 133: what does value of '1' mean?

lib/extensions/permissions.js
```diff
@@ -103,24 +107,40 @@ async function setAccess (udid, bundleId, permissionsMapping) {
...
+      'auth_value'
+    );
+  } catch {
```
mykola-mokhnach, line 136: shall we catch some particular error?

lib/extensions/permissions.js
KazuCocoa: Tried to find, but I was not able to figure `1` out yet. At least it is not in `YES|NO|limited|unset`

lib/extensions/permissions.js
KazuCocoa: Hm, I was only aware of no `auth_value` for the `allowed` query. When both are wrong, I expect the `allowed` one query will raise an error, then users may know something to report it

lib/extensions/permissions.js
KazuCocoa: `location=always|inuse|never|unset` did not match as well

PR #352, opened 2023-01-03
fix: retrieve an appropriate unix socket path

mwakizaka:
We have assumed that we always obtain `lsof -aUc launchd_sim` result in a specific order. (e.g. an output line like `/com.apple.CoreSimulator.SimDevice.D7082A5C-34B5-475C-994E-A21534423B9E` always comes before `com.apple.webinspectord_sim.socket` line) However, sometimes the assumption seems to be wrong (as far as I checked it can happen since Xcode 13.4.1. I haven't seen this phenomenon on Xcode 13.1 or earlier). Therefore, it is most likely necessary to obtain an appropriate unix socket path without depending on the order of the output.

### DIFF
--- a/lib/simulator-xcode-9.3.js
+++ b/lib/simulator-xcode-9.3.js
@@ -18,27 +18,8 @@ class SimulatorXcode93 extends SimulatorXcode9 {
       return this.webInspectorSocket;
     }
 
-    // lsof -aUc launchd_sim
-    // gives a set of records like:
-    //   COMMAND     PID      USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
-    //   launchd_s 81243 mwakizaka    3u  unix 0x9461828ef425ac31      0t0      /private/tmp/com.apple.launchd.ULf9wKNtd5/com.apple.webinspectord_sim.socket
-    //   launchd_s 81243 mwakizaka    4u  unix 0x9461828ef425bc99      0t0      /tmp/com.apple.CoreSimulator.SimDevice.F1191A22-11DD-408E-8CAF-0BC4A8F79E3B/syslogsock
-    //   launchd_s 81243 mwakizaka    6u  unix 0x9461828ef27d4c39      0t0      ->0x9461828ef27d4b71
-    //   launchd_s 81243 mwakizaka    7u  unix 0x9461828ef425dd69      0t0      ->0x9461828ef27d5021
-    //   launchd_s 81243 mwakizaka    8u  unix 0x9461828ef425b4c9      0t0      /private/tmp/com.apple.launchd.88z8qTMoJA/Listeners
-    //   launchd_s 81243 mwakizaka    9u  unix 0x9461828ef425be29      0t0      /private/tmp/com.apple.launchd.rbqFyGyXrT/com.apple.testmanagerd.unix-domain.socket
-    //   launchd_s 81243 mwakizaka   10u  unix 0x9461828ef425b4c9      0t0      /private/tmp/com.apple.launchd.88z8qTMoJA/Listeners
-    //   launchd_s 81243 mwakizaka   11u  unix 0x9461828ef425c081      0t0      /private/tmp/com.apple.launchd.zHidszZQUZ/com.apple.testmanagerd.remote-automation.unix-domain.socket
-    //   launchd_s 81243 mwakizaka   12u  unix 0x9461828ef425def9      0t0      ->0x9461828ef425de31
-    //   launchd_s 35621 mwakizaka    4u  unix 0x7b7dbedd6d63253f      0t0      /tmp/com.apple.CoreSimulator.SimDevice.150983FD-82FB-4A7B-86DC-D3D264DD90E5/syslogsock
-    //   launchd_s 35621 mwakizaka    5u  unix 0x7b7dbedd6d62f727      0t0      /private/tmp/com.apple.launchd.zuM1XDJcwr/com.apple.webinspectord_sim.socket
-    //   launchd_s 35621 mwakizaka    9u  unix 0x7b7dbedd6d632607      0t0      /private/tmp/com.apple.launchd.KbYwOrA36E/Listeners
-    //   launchd_s 35621 mwakizaka   10u  unix 0x7b7dbedd6d62f727      0t0      /private/tmp/com.apple.launchd.zuM1XDJcwr/com.apple.webinspectord_sim.socket
-    //   launchd_s 35621 mwakizaka   11u  unix 0x7b7dbedd6d62e6bf      0t0      /private/tmp/com.apple.launchd.7wTVfXC9QX/com.apple.testmanagerd.unix-domain.socket
-    //   launchd_s 35621 mwakizaka   12u  unix 0x7b7dbedd6d632607      0t0      /private/tmp/com.apple.launchd.KbYwOrA36E/Listeners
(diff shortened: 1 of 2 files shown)


lib/simulator-xcode-9.3.js
```diff
@@ -22,30 +20,45 @@ class SimulatorXcode93 extends SimulatorXcode9 {
...
-      if (!record.includes(this.udid)) {
-        continue;
-      }
-      const match = WEBINSPECTOR_SOCKET_REGEXP.exec(record);
-      if (!match) {
-        return null;
-      }
-      this.webInspectorSocket = match[1];
-      return this.webInspectorSocket;
+    //   COMMAND     PID      USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
```
mykola-mokhnach, line 23: would it be possible to cover this logic with at least one simple unit test?

lib/simulator-xcode-9.3.js
```diff
@@ -22,30 +20,45 @@ class SimulatorXcode93 extends SimulatorXcode9 {
...
+    //   launchd_s 35621 mwakizaka   16u  unix 0x7b7dbedd6d62e84f      0t0      /private/tmp/com.apple.launchd.g7KQlSsvXT/com.apple.testmanagerd.remote-automation.unix-domain.socket
+    // these _appear_ to always be grouped together by PID for each simulator.
+    // Therefore, by obtaining simulator PID with an expected simulator UDID,
+    // we can get the correct `com.apple.webinspectord_sim.socket`
+    // without depending on the order of `lsof -aUc launchd_sim` result.
+    const {stdout} = await exec('lsof', ['-aUc', 'launchd_sim']);
+    const udidPattern = `([0-9]{1,5}).+${this.udid}`;
+    const udidMatch = stdout.match(new RegExp(udidPattern));
+    if (!udidMatch) {
+      log.debug(`Failed to get Web Inspector socket with ${udidPattern} pattern`);
```
mykola-mokhnach, line 50: perhaps it would make more sense to log the lsof output instead for debug purposes

lib/simulator-xcode-9.3.js
```diff
@@ -22,30 +20,45 @@ class SimulatorXcode93 extends SimulatorXcode9 {
...
+      return null;
     }
 
-    return null;
+    const pidPattern = `${udidMatch[1]}.+\\s+(\\S+com.apple.webinspectord_sim.socket)`;
```
mykola-mokhnach, line 54: dots should also be escaped

lib/simulator-xcode-9.3.js
```diff
@@ -22,30 +20,45 @@ class SimulatorXcode93 extends SimulatorXcode9 {
...
+    const pidMatch = stdout.match(new RegExp(pidPattern));
+    if (!pidMatch) {
+      log.debug(`Failed to get Web Inspector socket with ${pidPattern} pattern`);
```
mykola-mokhnach, line 57: same as above

lib/simulator-xcode-9.3.js
mwakizaka: thank you for your advice, added as https://github.com/appium/appium-ios-simulator/pull/352/commits/c00901a9ddea178c5581a7a57d96d8cee3f17c59

lib/simulator-xcode-9.3.js
mwakizaka: thanks, I wasn't sure it was ok to add such large log. added as https://github.com/appium/appium-ios-simulator/pull/352/commits/08febf22e6f8bfb75565892320322c32d4571296

lib/simulator-xcode-9.3.js
mwakizaka: added as https://github.com/appium/appium-ios-simulator/commit/08febf22e6f8bfb75565892320322c32d4571296

lib/simulator-xcode-9.3.js
mwakizaka: I don't think we need it as ESLint says it is unnecessary.
Screenshot 2023-01-03 at 16 48 29

mykola-mokhnach: I assume it should be double-backslash

test/unit/simulator-specs.js
```diff
@@ -96,4 +97,49 @@ describe('simulator', function () {
...
+    ];
+
+    testParams.forEach(({udid, line, expected}) => {
+      it(`should find a Web Inspector socket when it appears at the ${line} line of grouped outputs`, async function () {
+        const xcodeVersion = {major: 9, versionString: '9.3.0'};
+        xcodeMock.expects('getVersion').atLeast(1).returns(B.resolve(xcodeVersion));
+
+        const sim = await getSimulator(udid);
+        const webInspectorSocket = await sim.getWebInspectorSocket();
+        webInspectorSocket.should.equal(expected);
```
mykola-mokhnach, line 141: should we also verify webInspectorSocket value and the fact it does not get assigned twice if the getter is called multiple times?

lib/simulator-xcode-9.3.js
mykola-mokhnach: I assume this long comments now could be reduced to a unit test link

lib/simulator-xcode-9.3.js
mwakizaka: sorry, I misunderstood. fixed as https://github.com/appium/appium-ios-simulator/pull/352/commits/eb8f11f9516246bd2f8df80463d567f9fb5b458d

test/unit/simulator-specs.js
mwakizaka: thanks, done as https://github.com/appium/appium-ios-simulator/pull/352/commits/427f452afd349ad30d304dbdf920fbb9fccfd2e1

lib/simulator-xcode-9.3.js
mwakizaka: thanks, done as https://github.com/appium/appium-ios-simulator/pull/352/commits/adfe006c4d53d71947bc3d03b01415768f8745fc

test/unit/simulator-specs.js
```diff
@@ -96,4 +97,59 @@ describe('simulator', function () {
...
+    });
+
+    const testParams = [
+      {udid: '0829568F-7479-4ADE-9E51-B208DC99C107', line: 'first', expected: '/private/tmp/com.apple.launchd.ULf9wKNtd5/com.apple.webinspectord_sim.socket'},
+      {udid: 'B5048708-566E-45D5-9885-C878EF7D6D13', line: 'second', expected: '/private/tmp/com.apple.launchd.zuM1XDJcwr/com.apple.webinspectord_sim.socket'},
+    ];
+
+    testParams.forEach(({udid, line, expected}) => {
+      it(`should find a Web Inspector socket when it appears at the ${line} line of grouped records`, async function () {
+        const xcodeVersion = {major: 9, versionString: '9.3.0'};
```
mykola-mokhnach, line 136: this could be moved to beforeEach

mykola-mokhnach (approved):

test/unit/simulator-specs.js
mwakizaka: thank you, addressed as https://github.com/appium/appium-ios-simulator/pull/352/commits/303ea169cfe79c835b156b1ba03134dc34da5c40

PR #369, opened 2023-07-28
fix: Exclude empty entries from the list of system bundle identifiers

mykola-mokhnach:


lib/simulator-xcode-15.js
```diff
@@ -16,21 +16,27 @@ class SimulatorXcode15 extends SimulatorXcode14 {
       return this._systemAppBundleIds;
     }
 
-    const appsRoot = path.resolve(_.trim(await this.simctl.getEnv('IPHONE_SIMULATOR_ROOT')), 'Applications');
+    const appsRoot = path.resolve(
+      _.trim(await this.simctl.getEnv('IPHONE_SIMULATOR_ROOT')),
```
KazuCocoa, line 20: btw, do you remember if `this.simctl.getEnv` always has `\n` as the result (so it would be nice to trim inside the `getEnv`)? getEnv was defined so long ago, so i haven't added the trim in getEnv, but just wondered if other value via getEnv had `\n` in the past

KazuCocoa (approved):

lib/simulator-xcode-15.js
mykola-mokhnach: no idea. I assume we never had similar issues before because we did not need to use the actual string output of getEnv method 🤷

PR #388, opened 2023-09-25
fix: Only restart springboard if the system language is changed

mykola-mokhnach:
I assume there is no need to restart springboard if only keyboard or locale changes are applied. 

Relates to https://github.com/appium/appium/issues/19202

cc @mwakizaka

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -578,10 +578,15 @@ class SimulatorXcode9 extends SimulatorXcode8 {
       ])));
     }
 
-    // In order to translate com.apple.SpringBoard, system prompts for push notification and system prompts for location
-    await B.all(['com.apple.SpringBoard', 'com.apple.locationd'].map((arg) => this.simctl.spawnProcess([
-      'launchctl', 'stop', arg
-    ])));
+    if (globalPrefs.AppleLanguages) {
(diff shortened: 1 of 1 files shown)


KazuCocoa (approved):
makes sense to me

mwakizaka:
Thank you. It makes sense. Btw, how about checking `SBRecentLocale` value of `xcrun simctl spawn booted defaults read com.apple.springboard` as well? I mean, we don't need to call `xcrun simctl spawn booted launchctl stop com.apple.SpringBoard` if the SpringBoard is already translated into the specified language.
```
% xcrun simctl spawn booted defaults read com.apple.springboard
{
    ...
    SBParentalControlsMCContentRestrictions =     {
        countryCode = "";
    };
    SBRecentLocale = "ja-JP";
    ...
}
```

mykola-mokhnach:
> Thank you. It makes sense. Btw, how about checking `SBRecentLocale` value of `xcrun simctl spawn booted defaults read com.apple.springboard` as well? I mean, we don't need to call `xcrun simctl spawn booted launchctl stop com.apple.SpringBoard` if the SpringBoard is already translated into the specified language.

This would be even more sophisticated approach. My assumption is that if someone has already provided some language value then most likely they want it to change rather than keep as is.

PR #403, opened 2023-11-22
fix: Only restart system services if the language is actually changed

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/19440

Perhaps, we need to add a separate capability to xcuitest driver as the next step to control this behaviour, so users could choose what is more important for them: running/cached WDA or fully translated UI dialogs.

cc @mwakizaka

### DIFF
--- a/lib/simulator-xcode-9.js
+++ b/lib/simulator-xcode-9.js
@@ -23,6 +23,7 @@ const DOMAIN_KEYBOARD_PREFERENCES = 'com.apple.keyboard.preferences';
 // com.apple.locationd: translates system prompts for location
 // com.apple.tccd: translates system prompts for camera, microphone, contact, photos and app tracking transparency
 const SERVICES_FOR_TRANSLATION = ['com.apple.SpringBoard', 'com.apple.locationd', 'com.apple.tccd'];
+const GLOBAL_PREFS_PLIST = '.GlobalPreferences.plist';
 
 /**
  * Creates device and common Simulator preferences, which could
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
@mwakizaka Could you please help with testing of the PR?

lib/simulator-xcode-9.js
```diff
@@ -583,9 +595,21 @@ class SimulatorXcode9 extends SimulatorXcode8 {
     }
 
     if (globalPrefs.AppleLanguages) {
-      await B.all(SERVICES_FOR_TRANSLATION.map((arg) => this.simctl.spawnProcess([
-        'launchctl', 'stop', arg
-      ])));
+      if (previousAppleLanguages && previousAppleLanguages === globalPrefs.AppleLanguages) {
```
mwakizaka, line 598: Excuse me, I guess we should compare in string like 
`if (previousAppleLanguages && JSON.stringify(previousAppleLanguages) === JSON.stringify(globalPrefs.AppleLanguages)) {` because
I saw a WDA session was killed as [this gist](https://gist.github.com/mwakizaka/ab8bd5179aaed43be4b4cbd8453435d0) when I try `create session` -> `delete session` -> `create session`.

KazuCocoa: It looks like `_.isEqual(previousAppleLanguages, globalPrefs.AppleLanguages)`

mykola-mokhnach: thanks, fixed

PR #445, opened 2025-03-17
fix: Wait for springboard restart after applesimutils invocation

mykola-mokhnach:
Related to https://github.com/appium/appium/issues/21116

PR #447, opened 2025-03-30
fix: Wait for springboard restart and spotlight application restart after applesimutils invocation

KazuCocoa:
Closes https://github.com/appium/appium/issues/21116.
Inherits https://github.com/appium/appium-ios-simulator/pull/445.

### DIFF
--- a/lib/extensions/permissions.js
+++ b/lib/extensions/permissions.js
@@ -287,11 +287,15 @@ async function runAndWaitForSystemReadiness(fn, timeoutMs) {
     // Make sure the springboard process restarted first.
     const timer = new timing.Timer().start();
     await waitForNewPid(initialSpringboardPid, SPRINGBOARD_BUNDLE_ID, timeoutMs);
-    const remainingTimeout = timeoutMs - timer.getDuration().asMilliSeconds;
+    const remainingTimeoutMs = timeoutMs - timer.getDuration().asMilliSeconds;
+    if (remainingTimeoutMs <= 0) {
+      // no need to check the SPOTLIGHT_BUNDLE_ID
+      return [false, result];
+    }
 
     // Then, checking if the new spring board process refreshes applications.
     // Spotlight.app is widely used so the app process can be an indicator to check the refresh.
-    await waitForNewPid(initialSpotlightPid, SPOTLIGHT_BUNDLE_ID, remainingTimeout);
+    await waitForNewPid(initialSpotlightPid, SPOTLIGHT_BUNDLE_ID, remainingTimeoutMs);
   } catch {
     return [true, result];
   }
--- a/lib/extensions/permissions.js
+++ b/lib/extensions/permissions.js
@@ -255,22 +255,22 @@ async function setAccess (bundleId, permissionsMapping) {
  * @returns {Promise<[boolean, T]>}
  */
 async function runAndWaitForSpringboardRestart(fn, timeoutMs) {
-  /** @type {() => Promise<number | undefined>} */
-  const getSpringboardPid = async () => {
+  /** @type {(string) => Promise<number | undefined>} */
+  const getPid = async (bundleId) => {
     try {
-      return (await this.ps()).find(({name}) => SPRINGBOARD_BUNDLE_ID === name)?.pid;
+      return (await this.ps()).find(({name}) => bundleId === name)?.pid;
     } catch {}
   };
 
-  /** @type {() => Promise<number | undefined>} */
-  const getSpotlightPid = async () => {
-    try {
-      return (await this.ps()).find(({name}) => SPOTLIGHT_BUNDLE_ID === name)?.pid;
-    } catch {}
+  const waitForNewPid = async (initialPid, bundleId) => {
+    await waitForCondition(async () => {
+      const pid = await getPid(bundleId);
+      return _.isInteger(pid) && initialPid !== pid;
+    }, {waitMs: timeoutMs, intervalMs: 500});
   };
 
-  const initialSpringboardPid = await getSpringboardPid();
-  const initialSpotlightPid = await getSpotlightPid();
+  const initialSpringboardPid = await getPid(SPRINGBOARD_BUNDLE_ID);
+  const initialSpotlightPid = await getPid(SPOTLIGHT_BUNDLE_ID);
   const result = await fn();
   if (!_.isInteger(initialSpringboardPid)) {
     // there is no point to wait if springboard was not running before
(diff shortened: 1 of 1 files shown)


lib/extensions/permissions.js
```diff
@@ -215,16 +220,74 @@ async function setAccess (bundleId, permissionsMapping) {
...
 
+/**
+ *
+ * @this {import('../types').CoreSimulator}
+ * @template {any} T
+ * @param {() => Promise<T>} fn
+ * @param {number} timeoutMs
+ * @returns {Promise<[boolean, T]>}
+ */
+async function runAndWaitForSpringboardRestart(fn, timeoutMs) {
```
mykola-mokhnach, line 257: the name of this function might be confusing as it also restarts the spotlight service.

Perhaps, we could rename it to something more specific, like runAndWaitForSystemReadiness

lib/extensions/permissions.js
```diff
@@ -215,16 +220,74 @@ async function setAccess (bundleId, permissionsMapping) {
...
+  };
+
+  const waitForNewPid = async (initialPid, bundleId) => {
+    await waitForCondition(async () => {
+      const pid = await getPid(bundleId);
+      return _.isInteger(pid) && initialPid !== pid;
+    }, {waitMs: timeoutMs, intervalMs: 500});
+  };
+
+  const initialSpringboardPid = await getPid(SPRINGBOARD_BUNDLE_ID);
```
mykola-mokhnach, line 272: I don't see a need to have getPid called twice. We may improve this helper to accept array of bundle identifiers and returns matched pids in an array as well, since a single `ps` output anyway contains all items

lib/extensions/permissions.js
```diff
@@ -215,16 +220,74 @@ async function setAccess (bundleId, permissionsMapping) {
...
+  const initialSpotlightPid = await getPid(SPOTLIGHT_BUNDLE_ID);
+  const result = await fn();
+  if (!_.isInteger(initialSpringboardPid)) {
```
mykola-mokhnach, line 275: should we also check if spotlight pid is defined?

lib/extensions/permissions.js
```diff
@@ -215,16 +220,74 @@ async function setAccess (bundleId, permissionsMapping) {
...
+    // there is no point to wait if springboard was not running before
+    return [false, result];
+  }
+
+  try {
+    await waitForNewPid(initialSpringboardPid, SPRINGBOARD_BUNDLE_ID);
+    // To check if the new spring board pid refresh applications.
+    // Spotlight.app is widely used so the app process can be an indicator to check the refresh.
+    await waitForNewPid(initialSpotlightPid, SPOTLIGHT_BUNDLE_ID);
```
mykola-mokhnach, line 284: I just wanted to confirm the logic here.
Should we wait 15 sec max until springboard is restarted and then +15 second max for spotlight to be restarted
OR
should we wait max 15 sec for both services to be restarted?

lib/extensions/permissions.js
KazuCocoa: Perhaps good naming.

The springboard restart caused applications' restarts, such as `Spotlight.app`. So, I initially thought such application restarts can also be part of the springboard restart method. Then, this method is ready == other services/apps restarted by springboard are also ready.

lib/extensions/permissions.js
```diff
@@ -247,41 +247,51 @@ async function setAccess (bundleId, permissionsMapping) {
...
+  if (!_.isInteger(initialSpringboardPid) || !_.isInteger(initialSpotlightPid)) {
+    // there is no point to wait if relevant processes were not running before
     return [false, result];
   }
 
   try {
-    await waitForNewPid(initialSpringboardPid, SPRINGBOARD_BUNDLE_ID);
-    // To check if the new spring board pid refresh applications.
+    // Make sure the springboard process restarted first.
+    const startAt = new Date().getTime();
```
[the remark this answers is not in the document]
KazuCocoa: oh, didn't notice that

lib/extensions/permissions.js
```diff
@@ -215,16 +220,84 @@ async function setAccess (bundleId, permissionsMapping) {
...
+        return false;
+      }
+    }, {waitMs: timeoutMs, intervalMs: 500});
+  };
+
+  let initialProcesses = [];
+  try {
+    initialProcesses = await this.ps();
+  } catch {};
+
```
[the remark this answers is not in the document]
KazuCocoa: This error is not important as this usage right now, so it's ok to suppress the error

lib/extensions/permissions.js
```diff
@@ -215,16 +220,84 @@ async function setAccess (bundleId, permissionsMapping) {
...
+    const shouldWaitForSystemReadiness = !_.isEmpty(
+      _.intersection(SERVICES_NEED_SPRINGBOARD_RESTART, _.keys(wixPermissions))
+    );
+    if (shouldWaitForSystemReadiness) {
+      const [didTimeout] = await runAndWaitForSystemReadiness.bind(this)(
+        execWixFn, SYSTEM_SERVICE_RESTART_TIMEOUT_MS
+      );
+      if (didTimeout) {
+        this.log.warn(
+          `The Springboard service and apps restarted by the springboard process restart did not restart ` +
```
mykola-mokhnach, line 237: The required system services did not restart after ...

lib/extensions/permissions.js
```diff
@@ -215,16 +220,84 @@ async function setAccess (bundleId, permissionsMapping) {
...
+  if (!_.isInteger(initialSpringboardPid) || !_.isInteger(initialSpotlightPid)) {
+    // there is no point to wait if relevant processes were not running before
+    return [false, result];
+  }
+
+  try {
+    // Make sure the springboard process restarted first.
+    const timer = new timing.Timer().start();
+    await waitForNewPid(initialSpringboardPid, SPRINGBOARD_BUNDLE_ID, timeoutMs);
+    const remainingTimeout = timeoutMs - timer.getDuration().asMilliSeconds;
```
mykola-mokhnach, line 290: remainingTimeoutMs

lib/extensions/permissions.js
```diff
@@ -215,16 +220,84 @@ async function setAccess (bundleId, permissionsMapping) {
...
+
+    // Then, checking if the new spring board process refreshes applications.
+    // Spotlight.app is widely used so the app process can be an indicator to check the refresh.
+    await waitForNewPid(initialSpotlightPid, SPOTLIGHT_BUNDLE_ID, remainingTimeout);
```
mykola-mokhnach, line 294: we may shortcircuit here if remainingTimeout <= 0

lib/extensions/permissions.js
KazuCocoa: done

lib/extensions/permissions.js
KazuCocoa: done

lib/extensions/permissions.js
KazuCocoa: done. actially this case could be rare as previous wait for the thing goes to catch clause then but it's reasonable to add the condition check

lib/extensions/permissions.js
```diff
@@ -215,16 +220,88 @@ async function setAccess (bundleId, permissionsMapping) {
...
+  }
+
+  try {
+    // Make sure the springboard process restarted first.
+    const timer = new timing.Timer().start();
+    await waitForNewPid(initialSpringboardPid, SPRINGBOARD_BUNDLE_ID, timeoutMs);
+    const remainingTimeoutMs = timeoutMs - timer.getDuration().asMilliSeconds;
+    if (remainingTimeoutMs <= 0) {
+      // no need to check the SPOTLIGHT_BUNDLE_ID
+      return [false, result];
```
mykola-mokhnach, line 293: return [true, result];

mykola-mokhnach (approved):

lib/extensions/permissions.js
KazuCocoa: oh, yea

PR #500, opened 2026-07-30
chore: Drop esmock

mykola-mokhnach:


test/unit/simulator.spec.ts
```diff
@@ -19,25 +18,35 @@ let currentGetVersion: (...args: any[]) => any = async () => ({
...
-      getVersion: (...args: any[]) => currentGetVersion(...args),
-    },
-    '../../lib/utils/xcode.js': {
-      assertXcodeVersion: (...args: any[]) => currentAssertXcodeVersion(...args),
-    },
-    '../../lib/utils/get-devices.js': {
-      getDevices: (...args: any[]) => currentGetDevices(...args),
-    },
+mock.module('teen_process', {
+  exports: {
```
eglitise: Should all of these be `namedExports` instead? https://github.com/appium/appium-adb/pull/898/changes/faf3dc1a5849e37e54a5d094e99d8e908909bd29..57ea0bf0bdd71fc07856410688ecadef8f57836c

test/unit/simulator.spec.ts
```diff
@@ -19,25 +18,35 @@ let currentGetVersion: (...args: any[]) => any = async () => ({
...
+    spawn: teenProcess.spawn,
+    SubProcess: teenProcess.SubProcess,
```
eglitise, line 24: Why not simply `...teenProcess`?
Same in `utils.spec.ts`

test/unit/simulator.spec.ts
mykola-mokhnach: yes it should

test/unit/simulator.spec.ts
mykola-mokhnach: because we don't need to mock all properties, only this one

eglitise (approved):

PR #502, opened 2026-09-14
refactor: mix in extension methods via prototype instead of class fields

mykola-mokhnach:
Extension methods were assigned as class fields, which become own instance properties rather than prototype methods, so subclass overrides couldn't call super.methodName(). Mirrors appium-base-driver's approach: functions stay on the extension modules and get Object.assign'd onto the prototype, with declare module augmentations providing the types.


eglitise (approved):
Interesting that SimulatorXcode14 did not implement SupportsAppPermissions

mykola-mokhnach:
> Interesting that SimulatorXcode14 did not implement SupportsAppPermissions

I anyway wanted to drop it soon and only keep Xcode 15 as the minimum.
But I will merge that pr with the simctl -> coresim replacement later

PR #503, opened 2026-09-16
feat!: replace node-simctl and applesimutils with @appium/coresim 

mykola-mokhnach:
## Summary

Replaces `node-simctl`, `applesimutils`, and `sqlite3`/`lsof` shell-outs with direct calls to `@appium/coresim`, a native addon that drives `CoreSimulator.framework` in-process instead of shelling out to `simctl`. Also drops Xcode 14 support and introduces an abstract `BaseSimulator` class for `instanceof` checks.

### Native migration
- All device/app lifecycle (boot/shutdown/erase/create/delete/install/launch/terminate/stat) now goes through `@appium/coresim` instead of `node-simctl`.
- TCC permission grant/revoke/reset/read now use native TCC.db writes via coresim (empirically confirmed to not kill the target app process, matching `applesimutils`' behavior) instead of `applesimutils` + raw `sqlite3` reads. `location`/`location-always` remain on `xcrun simctl privacy` directly (the one privacy action CoreSimulator doesn't model as a plain TCC row) — via `teen_process`, not `node-simctl`.
- Guest-side `launchctl`/`defaults` calls and Web Inspector socket discovery are now native calls instead of CLI-mediated (`simctl spawn`) or `lsof`-based ones.
- Biometric enrollment/matching uses coresim's native Darwin-notification wrappers instead of `notifyutil`.
- `node-simctl` and `applesimutils` are no longer dependencies.

### New public `Simulator` methods (backed by coresim)
`getAppContainer`, `appInfo`, `getPasteboard`/`setPasteboard`, `getScreenshot`, `addMedia`, `spawnProcess` (a generic guest-process escape hatch).

### Xcode 14 deprecated
`lib/simulator-xcode-14.ts` is removed; Xcode 15 is now the minimum supported version (`MIN_SUPPORTED_XCODE_VERSION = 15`). Its core lifecycle logic is merged into `SimulatorXcode15`, which now directly implements what used to be Xcode-15-only overrides (system-app-aware `isAppInstalled`, `getLaunchDaemonsRoot`, native UI appearance/contrast/content-size) as the single implementation, eliminating the dead "Xcode too old" throwing code path. `SimulatorXcode27` is unaffected.

### BaseSimulator
Adds an abstract `BaseSimulator` (in `lib/base-simulator.ts`), a minimal `EventEmitter` subclass with no behavior of its own, exported from the package root as the generic `instanceof` marker for "is this a Simulator" checks — every concrete Simulator class extends it.

BREAKING CHANGE: `Simulator`/`CoreSimulator` no longer exposes a `simctl` property (a raw `node-simctl` `Simctl` instance).
BREAKING CHANGE:  Constructing a Simulator for Xcode < 15 now throws.
BREAKING CHANGE:  `SimulatorXcode14` no longer exists. `BaseSimulator` is exported in its place for `instanceof` checks (e.g. replace `'simctl' in device` with `device instanceof BaseSimulator`).


eglitise:
If `simctl` is removed, how would this affect the XCUITest driver's `mobile: simctl` execute method? Is all of its supported functionality still supported here?

mykola-mokhnach:
> If `simctl` is removed, how would this affect the XCUITest driver's `mobile: simctl` execute method? Is all of its supported functionality still supported here?

xcuitest can (and will) support simctl directly. It's just that we don't need to expose it or use it in this particular package

lib/extensions/permissions.ts
```diff
@@ -100,13 +99,13 @@ export async function getPermission(
...
+function toPermissionService(serviceName: string): SimPermissionService {
   const lowerName = serviceName.toLowerCase();
-  if (Object.hasOwn(SERVICES, lowerName)) {
-    return SERVICES[lowerName as keyof typeof SERVICES] as string;
+  if ((SERVICES as readonly string[]).includes(lowerName)) {
+    return lowerName as SimPermissionService;
   }
   throw new Error(
-    `'${serviceName}' is unknown. Only the following service names are supported: ${JSON.stringify(Object.keys(SERVICES))}`,
+    `'${serviceName}' is unknown. Only the following service names are supported: ${JSON.stringify(SERVICES)}`,
```
KazuCocoa, line 114: [P1] Preserve the previously supported permission services

The new whitelist rejects `notifications`, `userTracking`, and `faceid`, which the previous AppleSimulatorUtils-backed setter supported. I reproduced this against an iOS 26.5 simulator: `setPermission(bundleId, service, 'yes')` throws an `is unknown` error for each of these services before reaching the native implementation. Existing callers that configure these permissions through `setPermission` or `setPermissions` will now fail. Could we retain support for these services through appropriate native implementations and add regression coverage?

lib/extensions/permissions.ts
```diff
@@ -204,63 +172,89 @@ async function setAccess(
...
+  const service = toPermissionService(serviceName);
+  switch (formatStatus(status).toLowerCase()) {
+    case STATUS.YES:
+      return await this._native.grantPermission(this.udid, service, bundleId);
+    case STATUS.NO:
+      return await this._native.revokePermission(this.udid, service, bundleId);
+    case STATUS.UNSET:
+      return await this._native.resetPermission(this.udid, service, bundleId);
+    default:
+      throw this.log.errorWithException(`'${status}' is not a supported value for '${serviceName}'`);
```
KazuCocoa, line 265: Is dropping support for `photos: 'limited'` an intentional change?

The previous setter forwarded `limited` to AppleSimulatorUtils, which supports this value for Photos on iOS 14 and later. This switch now only accepts `yes`, `no`, and `unset`; I confirmed that `setPermission(bundleId, 'photos', 'limited')` throws `'limited' is not a supported value for 'photos'`, while the getter still handles the limited state. If this removal is intentional, could we explicitly document it as a breaking change and describe the supported alternative? Otherwise, we should preserve the limited-access setter and add a round-trip test.

lib/extensions/permissions.ts
```diff
@@ -115,46 +114,17 @@ function formatStatus(status: string): string {
...
-      cause: e,
-    });
-  }
+async function execSimctlPrivacy(
+  this: CoreSimulatorWithAppPermissions,
+  action: 'grant' | 'revoke' | 'reset',
+  service: string,
+  bundleId: string,
+): Promise<void> {
+  await exec('xcrun', ['simctl', 'privacy', this.udid, action, service, bundleId]);
```
KazuCocoa: [P2] Forward `devicesSetPath` to the privacy command

This direct `simctl` invocation omits `--set`, so it always searches the default device set even when the Simulator instance targets a custom set. I created and booted an iPhone 17 / iOS 26.5 in an isolated set: `sim.setPermission('com.apple.mobilesafari', 'location', 'yes')` failed with exit code 148, while the same `simctl privacy` command succeeded after adding `--set <customSetPath>`. The old node-simctl wrapper forwarded this option automatically. Please include it when `this.devicesSetPath` is set; `location-always` uses the same affected path. A booted custom-set permission test would catch this, whereas the current custom-set tests only exercise lookup and `stat()`.

KazuCocoa (commented):
Reviewed 87916f0 and exercised the changes on actual simulators. The build, lint, formatting check, and all 36 unit tests passed. All 41 functional tests passed on both Xcode 26.6 / iOS 26.5 and Xcode 27 beta / iOS 27.0. Additional compatibility checks identified the permission issues below, which the current functional suite does not cover.

lib/extensions/permissions.ts
mykola-mokhnach: Fixed in eb0ee38 — `execSimctlPrivacy` now forwards `--set <devicesSetPath>` to `xcrun simctl privacy` when set, and there's a new regression test that boots a custom-set device and exercises `location` through it.

lib/extensions/permissions.ts
mykola-mokhnach: Fixed in eb0ee38, using @appium/coresim 1.2.0's expanded permission support: `faceid` and `userTracking` are now backed by native TCC rows again (same mechanism as the other services), with regression tests. `notifications` remains dropped — it was never a plain TCC row; the old AppleSimulatorUtils setter wrote a hand-built legacy bplist into BulletinBoard/SectionInfo.plist (marked "Legacy"/"Xcode 9 support" in that project's own source) with no confirmed modern equivalent, so this is a deliberate, documented breaking change.

lib/extensions/permissions.ts
mykola-mokhnach: Fixed in eb0ee38, using @appium/coresim 1.2.0's new 'limited' status on grantPermission — `setPermission(bundleId, 'photos', 'limited')` is supported again (exclusively for photos, matching the old behavior), with a round-trip regression test.

KazuCocoa (approved):

PR #504, opened 2026-09-18
fix: match simulator udid lookups case-insensitively

mykola-mokhnach:
getSimulatorInfo(), getSimulator(), and stat()/deviceState() compared a caller-supplied udid against CoreSimulator's own reported (canonically uppercased) udid with strict equality. getSimulator() now also canonicalizes the udid it constructs a Simulator with, once a case-insensitive match is found, so downstream native command dispatch and device-state checks agree with CoreSimulator's own case even when the original caller didn't.


lib/utils/devices.ts
```diff
@@ -11,7 +11,8 @@ export async function getSimulatorInfo(
   opts: SimulatorInfoOptions = {},
 ): Promise<SimulatorListEntry | undefined> {
   const devices = await listSimulators(opts);
-  return devices.find((sim) => sim.udid === udid);
+  const normalizedUdid = String(udid).toLowerCase();
```
eglitise, line 14: What's the purpose of explicitly casting this as a string, if the udid is already expected to be one?

mykola-mokhnach: just to avoid an exception if someone decides to put something else there

eglitise (approved):

